PR #25, opened 2020-10-28
feat: multitenancy

zachdaniel:
### Contributor checklist


priv/test_repo/migrations/20201028042738_migrate_resources2.exs
```diff
@@ -0,0 +1,20 @@
...
+  Updates resources based on their most recent snapshots.
+
+  This file was autogenerated with `mix ash_postgres.generate_migrations`
+  """
+
+  use Ecto.Migration
+
+  def up() do
+    create table(:multitenant_orgs, primary_key: false) do
+      add(:id, :binary_id, null: true, default: fragment("uuid_generate_v4()"), primary_key: true)
```
kingshalaby1, line 12: `null: false` ?
I noticed that IDs allow `nil` in many places, is that intended?

kingshalaby1:
this PR solves bug #26 , we can close it after merge

priv/test_repo/migrations/20201028042738_migrate_resources2.exs
zachdaniel: Really its just up to the resource wether or not its possible to be null, but it wouldn't really make sense in most contexts

PR #37, opened 2021-01-09
Omit field opts if they are default values

zimt28:
This works fine but the implementation isn't too nice. Alternative ideas?


zachdaniel:
TBH, the implementation seems reasonable to me :)

Just need to update some of the tests that are expecting those to be specified.

zachdaniel:
Looks great!

zachdaniel:

PR #38, opened 2021-01-10
Use ecto_sql formatter settings

zimt28:
Use `ecto_sql`'s formatter config when generating migrations. 

I'm using `File.cwd!/1` here, as suggested by José:
> if the project is being compiled with Mix, "File.cwd!" will always point to the root directory because Mix always runs from root
(https://groups.google.com/g/elixir-lang-talk/c/Ls0eJDdMMW8/m/VLWWAKWPAQAJ)

### DIFF
--- a/lib/migration_generator/migration_generator.ex
+++ b/lib/migration_generator/migration_generator.ex
@@ -509,7 +509,7 @@ defmodule AshPostgres.MigrationGenerator do
 
   defp format(string, opts) do
     if opts.format do
-      Code.format_string!(string)
+      Code.format_string!(string, locals_without_parens: ecto_sql_locals_without_parens())
     else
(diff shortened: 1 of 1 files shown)


zimt28:
I'd also like to keep the resource's attribute order for the migration but haven't figured yet out how the sorting currently works :)

zachdaniel:
@zimt28 sorting is complicated because there are order dependent operations. So we have a bunch of fall-through cases for determining if one operation has to happen after another. So what you could do is add a case at the bottom of that for the `AddAttribute`, `RemoveAttribute` and `AlterAttribute` to sort them based on their position in the resource's attributes.

zachdaniel:
More great work :)

PR #44, opened 2021-01-12
Update primary keys & formatting

zimt28:


zimt28:
Mhh need to update some tests. Strangely on my machine other tests are failing 🤔

.formatter.exs
```diff
@@ -12,6 +12,7 @@ locals_without_parens = [
 ]
 
 [
+  import_deps: [:ash],
```
zachdaniel, line 15: We can't (currently) modify the `.formatter` in the extensions (so the test files have to stay unformatted). This is because their formatter file is generated and confirmed.

zimt28: Interesting, didn't know that. Where does it get generated? I will send a new tomorrow which just changes the primary keys, there are probably a lot of failing tests because of the new check (`"Primary keys must not be allowed to be nil"`)

zimt28: Nevermind, missed the comment at the top

PR #47, opened 2021-01-27
Preserve attribute order

zimt28:


zimt28:
Will fix the tests later. I just covered new attributes for now, would this:
```elixir
defp after?(
         %_{attribute: %{order: l}},
         %_{attribute: %{order: r}}
       ),
       do: l > r
```
work for other operations as well?

zachdaniel:
That should work, yes. The interesting thing there is that `alter_attribute` uses `old_attribute` and `new_attribute`, so what order should that be in? The order of the old attribute or the order of the new attribute?

Might need to do something like this to account for that:
```elixir
defp after?(
         %_{old_attribute: %{order: l}},
         %_{attribute: %{order: r}}
       ),
       do: l > r
```

zachdaniel:
@zimt28 I wonder if removing the sort by name that I added is actually all we need. The default sorting unless the migration had to sort it differently for some reason would then be in the proper order, right?

zimt28:
That was the first thing I tried and it didn't work. The `Enum.group_by(& &1.name)` in `merge_attributes` returns a map, so the order isn't preserved there. 

I will however change the place where the index gets added:
```diff
  defp attributes(resource) do
    repo = AshPostgres.repo(resource)

    resource
    |> Ash.Resource.attributes()
-   |> Enum.sort_by(& &1.name)
    |> Enum.map(&Map.take(&1, [:name, :type, :default, :allow_nil?, :generated?, :primary_key?]))
+   |> Enum.with_index()
+   |> Enum.map(fn {attr, i} -> Map.put(attr, :order, i) end)
    |> Enum.map(fn attribute ->
``` 

> The interesting thing there is that alter_attribute uses old_attribute and new_attribute, so what order should that be in? The order of the old attribute or the order of the new attribute?

Hmm, good question. All I have done so far is adding attributes, so I have to play with this a bit first .. But I think it's not even possible to cover all cases, as multiple resources with the same table will mess up the order anyway.

zachdaniel:
True. Either way, this addition will keep them in the same order where reasonable 👍

zachdaniel:
You'll have to add some logic to add an `order` field for snapshots that are older than this version (an annoyance that we'll always have to deal with around changes to what goes in the snapshots). Now that I think about it, we should probably store the `ash_postgres` version inside the snapshot itself, that way we can eventually just check that version.

PR #55, opened 2021-04-12
Relationships references should follow bigint ids schema as well 

WolfDan:
### Contributor checklist

On my previous PR I modified the integer primary keys to generate a `bigserial` id, while the test was successful I didn't take into account relationships which should be `bigint` as well, so this PR aims to solve that 

My worry here is the lack of tests regarding this, I'm not really familiar with the way you test things here @zachdaniel so a bit of help/guidance would be appreciated ^^/

### DIFF
--- a/lib/migration_generator/operation.ex
+++ b/lib/migration_generator/operation.ex
@@ -39,6 +39,9 @@ defmodule AshPostgres.MigrationGenerator.Operation do
     end
 
     def on_update(_), do: nil
+
+    def integer_to_bigint(:integer), do: ":bigint"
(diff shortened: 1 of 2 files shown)


zachdaniel:
Interesting, so I don't think this would be quite enough. We only want to make the integer a bigint if the other end would also be a bigint. But there are two points worth mentioning:

1.) we should just default `:integer` to `:bigint` as well, that is a better default.
2.) we should actually just omit the type in `references(...)` because the type needs to be the same, and that is the default behavior.

zachdaniel:
I'll go ahead and make those changes.

PR #63, opened 2021-04-28
improvement: Preserve attribute order

zimt28:
A new PR as I messed up the Git history on the old one. I've deleted the `priv/` folder and run the migrations, but I'm still getting this:

```elixir
def up do
  create table(:posts, primary_key: false) do
    add :id, :uuid, null: false, default: fragment("uuid_generate_v4()"), primary_key: true
    add :title, :text
    add :score, :bigint
    add :public, :boolean
    add :category, :citext
    add :type, :text, default: "sponsored"
    add :price, :bigint
  end

  create constraint(:posts, :price_must_be_positive, check: "type = 'sponsored' AND price > 0")

  alter table(:posts) do
    add :decimal, :decimal, default: 0
    add :status, :status
  end
```

### DIFF
--- a/lib/migration_generator/migration_generator.ex
+++ b/lib/migration_generator/migration_generator.ex
@@ -213,9 +213,7 @@ defmodule AshPostgres.MigrationGenerator do
   end
 
   defp add_order_to_operations({snapshot, operations}) do
-    %{attributes: attributes} = snapshot
-
-    operations_with_order = Enum.map(operations, &add_order_to_operation(&1, attributes))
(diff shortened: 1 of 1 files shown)


zachdaniel:
Ah, I think we need to add logic to push constraints to the bottom as well.

zachdaniel:
but it looks like your PR is working as intended. Probably will have unexpected ordering when multiple resources uses the same table, but that doesn't really matter.

zachdaniel:
Great work!

PR #64, opened 2021-04-30
fix: fix fragment processing broken

basicBrogrammer:
Let me know what kind of testing needs to be done and I can see what I can do. 

### Contributor checklist

### DIFF
--- a/lib/functions/fragment.ex
+++ b/lib/functions/fragment.ex
@@ -20,10 +20,10 @@ defmodule AshPostgres.Functions.Fragment do
   def new([fragment | rest]) do
     split = split_fragment(fragment)
 
-    if Enum.count(split, &(&1 == "")) != length(rest) do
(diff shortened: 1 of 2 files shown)


zachdaniel:
I think a simple test in `filter_test.exs` that does something like `Ash.Query.filter(Post, fragment("? = ?", post.id, ^post.id)` should do it.

basicBrogrammer:
> I think a simple test in `filter_test.exs` that does something like `Ash.Query.filter(Post, fragment("? = ?", post.id, ^post.id)` should do it.

I couldn't do ID. I got an error about postgrex expecting a binary when it received a string version of the id. So I used the title

zachdaniel:
Ah, all good. In the future, you can do `type(^post.id, :uuid)`.

zachdaniel:

PR #65, opened 2021-05-05
improvement: document script to iterate migrations

frankdugan3:
Simply adding a script to the migrations docs that can help with iterating resource changes.

Open to any feedback on wording, etc.


frankdugan3:
Also, didn't notice it until just now, but apparently my editor did some automatic format cleanup too! Can undo that if you don't like it.

zachdaniel (approved):

PR #87, opened 2022-04-19
improvement: Add EctoMigrationDefault protocol and implement defaults

### LINKED ISSUE #86: Default values need a protocol
As noted in this issue: https://github.com/ash-project/ash/pull/310

Our default value handling is rudimentary.  What we do currently:

```elixir
defp default(%{default: value, type: type}, _) do
    case Ash.Type.dump_to_native(type, value) do
      {:ok, value} when is_struct(value) ->
        to_string(value)

      {:ok, value} ->
        inspect(value)

      _ ->
        "nil"
    end
  rescue
    _ ->
      "nil"
  end
```

But what we need to do is add a protocol like `AshPostgres.ToSql` that can be implemented for various types and structs.

joshprice:
- Integer
- BitString
- DateTime
- Date
- Time

Fixes #86

### Contributor checklist


zachdaniel:

PR #94, opened 2022-05-08
improvement: handle fallback ecto migration default elegantly

joshprice:
Instead of hard defaulting to a `nil` default value when there is no
available protocol implementation for the type, we provide a more
sensible default as a protocol fallback.

### Contributor checklist

### DIFF
--- a/lib/ecto_migration_default.ex
+++ b/lib/ecto_migration_default.ex
@@ -13,14 +13,14 @@ defimpl EctoMigrationDefault, for: Any do
 
       `#{inspect(value)}`
 
-    The default value in the migration will be set to `#{inspect(value)}`.
-    This may not be correct, depending on the type, so edit your migration accordingly.
+    The default value in the migration will be set to `nil` and you can edit
+    your migration accordingly.
 
     To prevent this warning, implement the `EctoMigrationDefault` protocol
     for the appropriate Elixir type in your Ash project.
     """)
 
-    inspect(value)
+    "nil # #{inspect(value)}"
   end
 end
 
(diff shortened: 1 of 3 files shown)


zachdaniel:
🤔 I'm not sure if we want to do this as it is very likely to produce invalid defaults. What if we prompted them? We do a similar pattern for potentially detected renames

joshprice:
How does the prompt work? Is that where it comments out the potentially problematic line and adds a comment for review? That would avoid the situation where you have potentially invalid defaults.

zachdaniel:
https://github.com/ash-project/ash_postgres/blob/main/lib/migration_generator/migration_generator.ex#L1510

Here are the places we do it. We use `Mix.shell().yes?/1` when we're getting a yes/no prompt, and `Mix.shell().prompt/1` otherwise.

joshprice:
Not sure a prompt makes sense here, what did you have in mind as the prompt options?

joshprice:
You're right that we should not produce potentially invalid values so think this might be a good compromise? Not sure whether those comments will be moved by the formatter though.

```elixir
add :product_code, :text, default: nil # {"xyz"}
```

joshprice:
Have abandoned the idea of injecting a comment since we can't guarantee that's the final clause and could comment out other options

zachdaniel:
Oh hey just a thought, we could do the comment if we insert the comment and then a newline :)

joshprice:
> Oh hey just a thought, we could do the comment if we insert the comment and then a newline :)

Is the `default: <default>` really guaranteed to be the last thing on the line?

zachdaniel:
well it would be guaranteed if, in the case we do a comment, we do it like this `default: "nil", #some comment\n` because the newline would happen after the comment.

PR #97, opened 2022-05-27
improvement: handle list migration defaults

joshprice:
### Contributor checklist

### DIFF
--- a/lib/ecto_migration_default.ex
+++ b/lib/ecto_migration_default.ex
@@ -48,6 +48,12 @@ defimpl EctoMigrationDefault, for: BitString do
   end
 end
 
+defimpl EctoMigrationDefault, for: List do
+  def to_default(value) do
+    inspect(value)
+  end
+end
+
 defimpl EctoMigrationDefault, for: DateTime do
   def to_default(value) do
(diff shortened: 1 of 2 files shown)


lib/ecto_migration_default.ex
```diff
@@ -48,6 +48,12 @@ defimpl EctoMigrationDefault, for: BitString do
   end
 end
 
+defimpl EctoMigrationDefault, for: List do
+  def to_default(value) do
+    inspect(value)
```
zachdaniel, line 53: I'm not sure that this is sufficient. What if it was a list of things that were not supported? i.e `default: [%{a: 1}]`. I think we'd need to do something like:
```elixir
"[" <> Enum.map_join(value, ", ", &EctoMigrationDefault.to_default/1) <> "]"
```

zachdaniel (changes requested):

lib/ecto_migration_default.ex
joshprice: Good point! Will add some handling for those cases

zachdaniel:
This has been open for a long time, I'm going to close it for now.

PR #105, opened 2022-08-15
fix: Squash dashes in migration names.

jimsynz:
I accidentally generated a migration with a dasherised name.  This fixes that.


zachdaniel:

zachdaniel:
I like it. Dasherized migration names "work" but they are ugly and we should just fix them like this :)

zachdaniel (approved):

PR #115, opened 2022-10-24
fix: && operator in expressions to point to ash_elixir_and

MrGrinst:
### Bug

The `&&` operator was being expanded to `ash_elixir_or` instead of `ash_elixir_and`. Fixed the source and test.

### Contributor checklist

### DIFF
--- a/test/calculation_test.exs
+++ b/test/calculation_test.exs
@@ -93,7 +93,7 @@ defmodule AshPostgres.CalculationTest do
       })
       |> Api.create!()
 
-    assert %{first_name_and_bob: "bob"} =
+    assert %{first_name_and_bob: "fred"} =
              Author
              |> Ash.Query.filter(id == ^author.id)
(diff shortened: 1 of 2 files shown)


test/calculation_test.exs
```diff
@@ -93,7 +93,7 @@ defmodule AshPostgres.CalculationTest do
       })
       |> Api.create!()
 
-    assert %{first_name_and_bob: "bob"} =
+    assert %{first_name_and_bob: "fred"} =
```
MrGrinst, line 96: I think this might actually not pass. I'm a little confused now how the test was initially passing.

MrGrinst: alright, fixed

zachdaniel:
uh...wow. Great catch. I think most people are using `or` and `and`, at least I assume as much because no one caught this up to now. Thanks!

zachdaniel:

PR #131, opened 2023-02-15
fix: Use UTC for default generated timestamps

sevenseacat:
> In cases where the database server is not set to UTC, this fixes timezone inconsistencies when adding timestamp columns to tables with existing data, that would then have their timestamps set to non-UTC times.

Something I noticed in dev - when adding `inserted_at`/`updated_at` timestamps to a table, it generated a migration that looked like this:

```
    alter table(:posts) do
      add :inserted_at, :utc_datetime_usec, null: false, default: fragment("now()")
      add :updated_at, :utc_datetime_usec, null: false, default: fragment("now()")
    end
```

I am in UTC+8, and I presume my dev database server is as well. So when I ran this migration at 12.01pm local time for an existing table that had data in it, the existing rows had their timestamps set to `2023-02-15 12:01:39.984484`.

However, when inserting records via Ash, the timestamps are converted to UTC before saving. I inserted a record via iex at 1.57pm local time, and the timestamp in the database was saved as `2023-02-15 05:58:04.952904`.

So I think the default value set for the column should be now in UTC, or `fragment("(now() AT TIME ZONE 'utc')")`, instead of just `now()` for consistency.

I tested locally by rolling back the generated migration, updating it to use:

```
    alter table(:posts) do
      add :inserted_at, :utc_datetime_usec,
        null: false,
        default: fragment("(now() AT TIME ZONE 'utc')")

      add :updated_at, :utc_datetime_usec,
        null: false,
        default: fragment("(now() AT TIME ZONE 'utc')")
    end
```

and ran it again, and the timestamps were set in UTC like I expected.

I don't know how one would go about writing a test for this!

### Contributor checklist

### DIFF
--- a/lib/migration_generator/migration_generator.ex
+++ b/lib/migration_generator/migration_generator.ex
@@ -2282,7 +2282,7 @@ defmodule AshPostgres.MigrationGenerator do
           ~S[fragment("uuid_generate_v4()")]
 
         default == (&DateTime.utc_now/0) ->
-          ~S[fragment("now()")]
+          ~S[fragment("(now() AT TIME ZONE 'utc')")]
 
         true ->
           "nil"


sevenseacat:
I'm not sure I follow your logic. Ecto doesn't know anything about my DB running in UTC+8 - it converts all timestamps to UTC, and by default they're stored as timestamp w/o timezone, so provided times are always converted to UTC before they hit the DB. This is only for when the app doesn't provide times - we should always generate them in UTC also, no?

zachdaniel:
🤔 Yeah, I guess the default we add into the migrations is meant to mimic the values we would have set as default if we set them via the app, you're right. My bad.

zachdaniel:
Only issue is that when we merge this everyone who generates migrations is going to get a bunch of migrations generated for all of their timestamps :)

zachdaniel:
They could just delete the migration if they don't want to make the change, but I guess it should be seen as a good thing that we're ensuring their database is more consistent.

jimsynz:
Yup. We can probably emit a warning when generating said migrations also

zachdaniel:
It would be pretty difficult for the migration generator to tell the difference between this change and any other default value change (the migration default is configurable).

jimsynz:
I suppose so

sevenseacat:
I don't really like the idea of all users getting a surprise migration upon getting a minor update - maybe leave it for now (for a larger release or something) because it's not a huge problem? I leave it in your capable hands :)

zachdaniel:
Yeah, I think the only thing we could really do is make it an opt-in config that shows a warning if its not set. Kind of a PITA but there isn't really another good way. Or we could let this PR sit here until a big release happens.

zachdaniel:
We're going to be releasing a breaking change as part of the 3.0 release, so let's just merge this and warn people about it. They can just delete the generated migration code if they don't like it :)

PR #132, opened 2023-02-20
improvement: Implement string_join expr

totaltrash:
`string_join` in the data layer (see https://github.com/ash-project/ash/pull/515)

I've dumped some tests in calculation_test.exs, but I wonder if that's the right place? Oh, also, I'm not sure about the params getting passed on to the next call to do_dynamic_expr, (bindings, embedded)? I C&P that without understanding what's going on there.

### Contributor checklist


zachdaniel:

PR #152, opened 2023-06-07
fix: interpolate table names with `inspect` in generated migrations

vonagam:
Currently in `MigrationGenerator.Phase` a table name interpolated as simply `:#{table}` for `Create.up`, `Alter.up/down`, but in `Create.down` it is done as `:#{inspect(table)}`. If a table name contains non-word letters then without that inspect call the result may end up being an invalid code - `:auth.tokens` (if the table name is "auth.tokens") versus legal `:"auth.tokens"`.

Seems like having non-word letters might be a bad idea, still, a simple fix.

### DIFF
--- a/lib/migration_generator/operation.ex
+++ b/lib/migration_generator/operation.ex
@@ -26,9 +26,13 @@ defmodule AshPostgres.MigrationGenerator.Operation do
     def in_quotes(nil), do: nil
     def in_quotes(value), do: "\"#{value}\""
 
+    def as_atom(value) when is_atom(value), do: Macro.inspect_atom(:remote_call, value)
+    # sobelow_skip ["DOS.StringToAtom"]
+    def as_atom(value), do: Macro.inspect_atom(:remote_call, String.to_atom(value))
+
     def option(key, value) do
       if value do
-        "#{key}: #{inspect(value)}"
+        "#{as_atom(key)}: #{inspect(value)}"
       end
     end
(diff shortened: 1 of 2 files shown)


zachdaniel:
I wonder…could we add a helper that will only do this if necessary? Actually, the formatter might do that for us automatically…

vonagam:
Formatter removes unneeded quotes right now.

vonagam:
Ah, there is `format: false` in those two tests. For it to pass either that should be removed/set to true, or assertion containing `:posts` should search for `:"posts"` instead.

zachdaniel:
I'm fine just adjusting the tests to to check for `:"..."`. We need `format: false` so that we can assert on things being present in tests without line breaks from the formatter making it harder.

vonagam:
Hm... I'm thinking. I see at least 30 other places where `:#{table}` is used. It seems like a simple helper func should be added. It can be as simple as `def to_atom(name), do: inspect(String.to_atom(name))` which would give `:name` for "name" and `:"api.name"` for "api.name". Is there an appropriate place for such function?

zachdaniel:
I'm not sure. It could go in `MigrationGenerator`, or there is a helpers module in the operations file.

zachdaniel (dismissed):

zachdaniel (approved):

zachdaniel:

PR #155, opened 2023-06-22
fix: fix Logger deprecations for elixir 1.15

ahey:
Use Logger.warning instead of Logger.warn, [which is deprecated as of elixir 1.15](https://github.com/elixir-lang/elixir/blob/v1.15/CHANGELOG.md#logger-2).

Logger.warning [is available since elixir 1.11](https://hexdocs.pm/logger/1.15.0/Logger.html#warning/2), and since [this project requires at least elixir 1.11](https://github.com/ash-project/ash_postgres/blob/main/mix.exs#L15), this is a safe change.

### DIFF
--- a/lib/ecto_migration_default.ex
+++ b/lib/ecto_migration_default.ex
@@ -13,7 +13,7 @@ defimpl EctoMigrationDefault, for: Any do
   require Logger
 
   def to_default(value) do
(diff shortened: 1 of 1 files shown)


jimsynz (approved):
perfect

zachdaniel:

PR #162, opened 2023-08-05
improvement: custom-extension implementation

moissela:
Adds the ability to define custom versioned postgres functions like **ash-functions** through **AshPostgres.CustomExtension** behaviour implementation.

With this feature Ash projects and extensions can simply define a module that implements the callbacks **install/1** and **uninstall/1** and require that the users only add the custom extension to Repo.


test/support/test_no_sandbox_repo.ex
```diff
@@ -8,7 +8,7 @@ defmodule AshPostgres.TestNoSandboxRepo do
   end
 
   def installed_extensions do
-    ["ash-functions", "uuid-ossp", "pg_trgm", "citext"] --
+    ["ash-functions", "uuid-ossp", "pg_trgm", "citext", AshPostgres.TestCustomExtension.extension()] --
```
zachdaniel, line 11: This is cool :) I think we should just have them put the module here, though, and if they do we will call `.extension()` on it :)

moissela: Hi @zachdaniel, good tip!
I've fixed that and the formatting / credo issues 🙂

zachdaniel:
Thanks for your contribution!

PR #167, opened 2023-09-21
fix: use `:wrap_list` type instead of custom validaitons

vonagam:
Adds ability to pass single atom/string to `index` dsl. (So maybe it is an improvement, not a fix, but it is really small, don't know.)

While at it replaced custom validations that are not needed anymore (and one of them contained outdated field name `manages_tenant`).


zachdaniel:
`thanks

PR #175, opened 2023-10-31
feat: Add support for nulls_distinct in custom indexes

sezaru:
This will add support for the `nulls_distinct` option for PG 15

More information in https://hexdocs.pm/ecto_sql/Ecto.Migration.html#index/3-the-nulls_distinct-option

### Contributor checklist


sezaru:
Not sure why the mix test step fails, it runs just fine locally and the logs don't seem to give me a good reason

jimsynz:
It's failing because there are migrations that should be generated.

zachdaniel:
So the fact that this generates migrations for existing unchanged things is problematic. To address this, when loading a snapshot, we need to set the `nulls_distinct?` option to `false` if the option is not set.

PR #179, opened 2023-11-17
fix: support new upsert field options

barnabasJ:
### Contributor checklist

I tried to add the new `{replace_all_ecxept, [fields]}` upsert_fields option directly in a create action and I always got a Enum Protocl not implemented error. I was able to fix it in my code by adding this. 

I copied some of the code from https://github.com/ash-project/ash/blob/1cb544ddfa8f058fe56007079a8aa3accd42e50d/lib/ash/changeset/changeset.ex#L771 to make it work.

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -1289,6 +1289,31 @@ defmodule AshPostgres.DataLayer do
       )
   end
 
+  defp upsert_fields(fields, _, _) when is_list(fields) do
+    fields
+  end
+
+  defp upsert_fields({:replace, fields}, _, _) do
+    fields
+  end
+
+  defp upsert_fields(:replace_all, resource, _) do
+    resource
+    |> Ash.Resource.Info.attributes()
+    |> Enum.map(fn %{name: name} -> name end)
+  end
+
+  defp upsert_fields({:replace_all_except, except_fields}, resource, _) do
+    resource
+    |> Ash.Resource.Info.attributes()
+    |> Enum.map(fn %{name: name} -> name end)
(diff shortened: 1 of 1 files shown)


lib/data_layer.ex
```diff
@@ -1289,6 +1289,31 @@ defmodule AshPostgres.DataLayer do
...
+  defp upsert_fields({:replace_all_except, except_fields}, resource, _) do
+    resource
+    |> Ash.Resource.Info.attributes()
+    |> Enum.map(fn %{name: name} -> name end)
+    |> Enum.reject(fn name -> name in except_fields end)
+  end
+
+  defp upsert_fields(nil, _, changesets) do
+    changesets |> Enum.flat_map(&Map.keys(&1.attributes)) |> Enum.uniq()
+  end
```
barnabasJ, line 1315: Not sure about this one. I wanted your opinion first.

After this is clear to me, I will also add some tests.

zachdaniel: This is interesting. This *should* be handled by the changeset logic before it gets here. I think we may need to fix this in core somewhere.

zachdaniel: Yeah, I think the `upsert_fields` option should be transformed to just be a list of fields by core before passing the option to the data layer.

zachdaniel: The problem is that it can also be provided after compile time as an option. So I think somewhere in bulk create logic we need to do the transformation. There should be similar logic in `Ash.Changeset`

barnabasJ:
Should be fixed by https://github.com/ash-project/ash/pull/777

PR #180, opened 2023-11-18
improvement: allow specifying multi-column foreign keys

rbino:
* improvement: add match_with option on references
* improvement: add match_type option on references

Submitting this as a draft to get some early feedback given it's my first contribution.

What still needs to be done:
- [x] Handle the case where `with` has to be added both due to multitenancy _and_ because it's explicitly defined
- [x] Cover those cases with tests
- [x] Add tests for the Transformer check
- [ ] Polymorphic variants of this (I'd be more inclined to leave this to a separate PR probably)

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -15,6 +15,8 @@ spark_locals_without_parens = [
   include: 1,
   index: 1,
   index: 2,
+  match_type: 1,
+  match_with: 1,
   message: 1,
   migrate?: 1,
   migration_defaults: 1,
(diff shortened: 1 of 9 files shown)


zachdaniel:
Looks great! 👍

rbino:
@zachdaniel if the polymorphic variants can be implemented in a separate PR this should be ready for review

zachdaniel:
Looks like some tasks to run:

`mix spark.formatter`
`mix spark.cheat_sheets`

as well as a credo check to resolve.

rbino:
Fixed all three things

zachdaniel:

PR #183, opened 2023-11-20
Add minimal failing test for #182

rbino:
See https://github.com/ash-project/ash_postgres/issues/182


zachdaniel:
Realized this is actually broken in core, not here. Thanks for the reproduction!

PR #187, opened 2023-12-19
fix: replace upsert field with source in EXCLUDED fragment

barnabasJ:
### Contributor checklist


lib/data_layer.ex
```diff
@@ -1377,12 +1377,27 @@ defmodule AshPostgres.DataLayer do
...
+    case resource
+         |> Ash.Resource.Info.attributes()
+         |> Enum.find(&(&1.name == field)) do
+      %{source: source} when not is_nil(source) ->
+        source
+
+      _ ->
+        field
+    end
+  end
```
barnabasJ, line 1399: Is there a better way to get the argument than looking through all the attributes?

zachdaniel: `Ash.Resource.Info.attribute(resource, field)` will get it 👍

zachdaniel (approved):

PR #195, opened 2024-01-22
fix: keep fields of `custom_index` in format that they were provided

vonagam:
Right now fields of `custom_index` are first transformed into atoms and then during migration generation into strings. 

I used a field named "order" and encountered an error because of this. Ecto generates different statements depending on whenever a field was provided as a string or as an atom:
```elixir
create index(:some_table, ["order"]) # invalid
create index(:some_table, [:order]) # valid
```
For atoms the resulting name will be in quotes, for strings they will be left as is (and it, for example, allows specifying things like `"name gin_trgm_ops"` for full-text search usage).

So it makes sense to keep user provided input as is and not convert it into atoms or strings.

Also made `RemoveCustomIndex` operation to use reverse direction of `AddCustomIndex` to avoid repetition.

### DIFF
--- a/lib/aggregate.ex
+++ b/lib/aggregate.ex
@@ -77,7 +77,7 @@ defmodule AshPostgres.Aggregate do
               first_relationship =
                 case Ash.Resource.Info.relationship(resource, first_relationship) do
                   nil ->
-                    raise "No such relationship for #{inspect(first_relationship)} aggregates #{inspect aggregates}"
+                    raise "No such relationship for #{inspect(first_relationship)} aggregates #{inspect(aggregates)}"
 
                   first_relationship ->
                     first_relationship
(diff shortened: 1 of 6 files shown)


zachdaniel:
I think you're going to have to handle this when loading snapshots and encoding snapshots as well. Once we load the json snapshot, they will always end up being strings.

zachdaniel:
Oh, I think it's solvable, we just need to do a bit of extra work. When encoding we should store it as `{"type": "atom" | "string", value: "value"}`, and then decode it appropriately.

vonagam:
Ok, implemented such thing.

zachdaniel:
awesome work! 🚀 Thank you for your contribution! 🚀

PR #211, opened 2024-02-20
fix: use proper tables in joins originating from polymorphic resource

vonagam:
`context` is taken from a root query. If a root query is of polymorphic resource then it will contain `data_layer.table`. It should not be propagated into context of a join subquery otherwise it will use a root resource table for a related resource.


zachdaniel (approved):

zachdaniel:

PR #222, opened 2024-03-20
fix: generate correct custom index name in down migration function

rapidfsub:
### Contributor checklist

before fix, it generates wrong index name in down migration, especially multitenant case.
i just made code same as up migration.


zachdaniel (approved):

zachdaniel:

PR #240, opened 2024-04-18
feat: test with calculations using `exists`

lukasender:
# Test Coverage

Additional test that only increases test coverage.

I encountered a weird behaviour in our code base where a `caluclation` using an `exists` as defined in this PR would not return `true` even if the relation exists.

I wrote a test case directly in `ash_postgres` trying to reproduce the behaviour. However, I could not reproduce it. The test case now only serves to increase the test coverage and proves that the features behaves as expected.


zachdaniel:

PR #263, opened 2024-05-03
chore: add failing test for timestamptz

frankdugan3:
### Contributor checklist

Good news: It's not the bulk atomics. Bad news: All updates shift the TZ for `:timestamptz` PG types.


zachdaniel:
Yknow what...I think we need a timestamp with time zone type in Ash core for this to work, actually. Because the basic issue is that it ends up casting a timestamp without timezone to a timestamp with time zone.

```elixir
SELECT (now()::timestamp)::timestamp with time zone at time zone 'EST'
```

That simulates the issue.

zachdaniel:
So TBH I don't think we should merge these tests, as this is the expected behavior, looking at it like this.

PR #277, opened 2024-05-11
docs: change 'Get Started' guide for Ash 3.0 compatibility

arcanemachine:
Just a couple changes I needed to make as I was following the tutorial:

- Use newer `ash_postgres` version since old version had dependency resolution errors

- Change path of Helpdesk resources to match the examples seen in the main Ash `Get Started` tutorial.

- Use `Ash.read/update/load` instead of `Helpdesk.Support.read/update/load` since the compiler was giving deprecation warnings

- Fix wording at the end to avoid using confusing double-negatives

I think my changes are accurate, but should probably be double checked by someone who is more familiar with Ash. Thanks.

### Contributor checklist

- [x] Bug fixes include regression tests (N/A)
- [x] Features include unit/acceptance tests (N/A)

### DIFF
--- a/documentation/tutorials/get-started-with-ash-postgres.md
+++ b/documentation/tutorials/get-started-with-ash-postgres.md
@@ -25,7 +25,7 @@ In this guide we will:
 
 Add the `:ash_postgres` dependency to your application
 
-`{:ash_postgres, "~> 1.3.6"}`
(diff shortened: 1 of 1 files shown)


zachdaniel (approved):

zachdaniel:

PR #288, opened 2024-05-16
test: add failing test for `changing_attributes` check for create

vonagam:
Based on code and previous behavior (before being switched to `filter` from `match`) `changing_attributes` should work with create actions. But currently it does not.

### DIFF
--- a/test/ash_postgres_test.exs
+++ b/test/ash_postgres_test.exs
@@ -12,7 +12,15 @@ defmodule AshPostgresTest do
     }
   end
 
-  test "filter policies are applied" do
+  test "filter policies are applied in create" do
+    assert_raise Ash.Error.Forbidden, fn ->
+      AshPostgres.Test.Post
+      |> Ash.Changeset.for_create(:create, %{title: "worst"})
+      |> Ash.create!()
+    end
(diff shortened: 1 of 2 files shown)


zachdaniel:
Changing the test to

```elixir
  test "filter policies are applied in create" do
    assert_raise Ash.Error.Forbidden, fn ->
      AshPostgres.Test.Post
      |> Ash.Changeset.for_create(:create, %{title: "worst"}, authorize?: true)
      |> Ash.create!()
    end
  end
```

makes it pass, so I don't have a reproduction of the issue. However, I did notice that if you were using the `:from` option in the policy, i.e `from: nil, to: "foo"` that it would have caused it to appear as `false`, so I'm going to amend the `changing_attributes` policy to consider all attributes as going from `nil` to `the value` in ash core.

zachdaniel:
This is because the ash_postgres domain has 

```elixir
  authorization do
    authorize(:when_requested)
  end

```

vonagam:
Ah, missed that. I was having that issue on Ash 2 so assumed that the issue was reproduced on Ash 3 when got the failing test.

PR #289, opened 2024-05-16
improvement: support `on_delete: :nilify` for specific columns

davidebriani:
Introduce the possibility to specify which columns of a foreign key should be set to NULL when the referenced resource is deleted: `on_delete: {:nilify, columns}`.
This provides more fine-grained control over the `on_delete: :nilify`  option for configuring the foreign key constraint, where the default behavior is to set all columns of the foreign key to `NULL`.

This is useful for example in cases where multitenancy is employed on resources, and a resource A has a `(tenant_id, resource_b_id)` foreign key that references the `tenant_id` and `id` columns of resource B: when the resource B is deleted, we'd like to nilify only the `resource_b_id` column of resource A and not its `tenant_id`.

Note: this functionality is supported by Ecto's `on_delete: {:nilify, columns}` option and ultimately by Postgres, although only from v15.0 onwards.

For details see:
- https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK
- https://hexdocs.pm/ecto_sql/Ecto.Migration.html#references/2-options

### Contributor checklist

### DIFF
--- a/documentation/dsls/DSL:-AshPostgres.DataLayer.md
+++ b/documentation/dsls/DSL:-AshPostgres.DataLayer.md
@@ -258,7 +258,7 @@ end
 
 | Name | Type | Default | Docs |
 |------|------|---------|------|
-| [`polymorphic_on_delete`](#postgres-references-polymorphic_on_delete){: #postgres-references-polymorphic_on_delete } | `:delete \| :nilify \| :nothing \| :restrict` |  | For polymorphic resources, configures the on_delete behavior of the automatically generated foreign keys to source tables. |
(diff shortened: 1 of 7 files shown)


rbino:
Note for when this is undrafted: the CI runs on Postgres 14 (which would not support this) and Postgres 15 (which would), I don't know if there are already tags to run tests only in of the two

zachdaniel:
Oh, interesting. Since this only affects the migration generator, I think it will be fine to add and people can just not use it if they aren't on a version that supports it, but we'll probably need to set some environment variable so you can skip the tests on pg 14.

davidebriani:
@rbino thanks for the advice!
I only added a single unit test about the generation of migration files, so it shouldn't involve interaction with the database.
@zachdaniel fingers crossed for the CI :crossed_fingers:

zachdaniel:
One more thing, please add some information about this to this guide: https://github.com/ash-project/ash_postgres/blob/main/documentation/topics/resources/references.md

davidebriani:
I added a commit to note down the available options on the documentation page.
@zachdaniel let me know if the wording isn't right.

zachdaniel (approved):

zachdaniel:

PR #292, opened 2024-05-17
chore: failing test for calculation ordered paginated relationship

rbino:
### Contributor checklist


rbino:
If it can help debuggin the root cause: removing [this](https://github.com/ash-project/ash/blob/9fc5ddfe2bd6ac6c4188d62b25c8075638a514e9/lib/ash/actions/read/read.ex#L179) makes the offset test pass, but makes the keyset test fail with:
```
     ** (Postgrex.Error) ERROR 42P18 (indeterminate_datatype) could not determine data type of parameter $1

         query: SELECT DISTINCT p0."id", s1."id", s1."is_active", s1."name", s1."organization_id", s1."__order__" FROM "posts" AS p0 INNER JOIN LATERAL (SELECT su0."id" AS "id", su0."is_active" AS "is_active", su0."name" AS "name", su0."organization_id" AS "organization_id", row_number() OVER "order" AS "__order__" FROM "public"."users" AS su0 INNER JOIN "public"."post_followers" AS sp1 ON sp1."follower_id" = su0."id" WHERE ($1 IS NULL) AND (su0."id"::uuid > $2::uuid) AND (sp1."post_id" = p0."id") WINDOW "order" AS (ORDER BY sp1."order"::bigint, su0."id") ORDER BY sp1."order"::bigint, su0."id" LIMIT $3) AS s1 ON TRUE WHERE (p0."type"::varchar = $4::varchar) AND (p0."id" = ANY($5)) ORDER BY s1."__order__"
```

zachdaniel:
That should be helpful, yes :)

Issue #305, opened 2024-05-27
`Identity.where` is not working


jechol:
**Describe the bug**
The Identity.where option is being ignored.

**To Reproduce**

For example,

```
  identities do
    identity :foo_bar, [:foo], where: :bar_is_ten
  end

  postgres do
    identity_wheres_to_sql [bar_is_ten: "bar = 10",  
                            foo_bar: "bar = 20"]
  end
```

This results in bar_is_ten being ignored, and it creates the following with where: "bar = 20" instead:

```
  create unique_index(:my_table, ["foo"], name: "foo_bar_index", where: "bar = 20")
```

**Expected behavior**

A migration with "bar = 10"

```
 create unique_index(:my_table, ["foo"], name: "foo_bar_index", where: "bar = 10")
```

zachdaniel:
The keyword lists are not merged, you're meant to put them all in one call to the option.

zachdaniel:
The way that you're writing it is incorrect. The key is the identity name and the value is the SQL where clause.

```elixir

identities do
    identity :foo_bar, [:foo], where: expr(bar == 10)
  end

  postgres do
    identity_wheres_to_sql [foo_bar: "bar = 10"]
  end

```

jechol:
@zachdaniel 

So, the expression `expr(bar == 10)` in the where clause is not actually being used, right?

zachdaniel:
It's not being used to generate the filter in Postgres cases, no. (Because we don't have an expr -> raw SQL converter yet), it's all Ecto expressions.

PR #306, opened 2024-05-27
identity_where_to_sql to find with identity.where, not identity.name

jechol:
### Contributor checklist

Fixes https://github.com/ash-project/ash_postgres/issues/305


zachdaniel:
The key is the identity name, so this is not correct unfortunately.

PR #312, opened 2024-05-31
fix custom-extension versions issue

moissela:
I believe that with the latest evolutions, the correct functionality of the custom extensions has been lost at some point.

The idea is to allow defining SQL functions on the database through the project's migrations, also allowing version tracking to enable their evolution.

With the current code, even if a new version of the extension is implemented, AshPostgres proposes the "current version" in the new migration, whereas it should provide the latest actual version.

With this commit, the function is correctly restored.


zachdaniel:
Custom extensions must present the *old* version. The proposed version is always the *newest* version. We provide the old version so that you can provide a minimal migration from the old version to the new version. Imagine that the user has migrated from 0 -> 1, and then two new versions are published. We only want to provide the functionality changes from 1 -> 4, not all of them. The way that its been adjusted is intentional and how it must work to allow extensions to incrementally migrate.

moissela:
Got it! 🙂

PR #324, opened 2024-06-12
Add a failing test for sorting on relationship

shamanime:
Tried to come up with this test as close to my scenario as possible.

I'm not even sure if the second assert works but it gives the error I'm seeing in our app.


zachdaniel:

PR #330, opened 2024-06-18
add test to show error with building a reference

barnabasJ:
doing an atomic change while having `require_atomic false` leads to an error while building the references for the update


zachdaniel:

PR #332, opened 2024-06-18
fix: ensure index keys are atoms in generated migrations

emadshaaban92:


zachdaniel:

PR #339, opened 2024-06-29
improvement: order keys in snapshot json

vonagam:
This makes order of keys in a snapshot stable. Easier for computer and human eye to diff.

Note: There is [`StableJason`](https://github.com/egze/stable_jason) (which does the same thing but because of need to support structs, which snapshot does not use, it instead encodes -> decodes -> orders -> encodes). Upside is that the change would have looked even smaller but don't know if a dependency is worth it in this case.


zachdaniel:

PR #368, opened 2024-08-12
Handle filter condition on create

maennchen:
I implemented it so that the single create returns the existing row.

Any row that was skipped because of the upsert filter clause will have the metadata entry `upsert_skipped` set.

Alternatively we could return `nil`, but I believe that might confuse the users of the function. With bulk creates it makes more sense to be missing.

I couldn't find specific tests for single create / single bulk create. So I added those as well.

This is a prerequisite to implement the syntax sugar from https://github.com/ash-project/ash/issues/1385

### Contributor checklist

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -2539,6 +2539,28 @@ defmodule AshPostgres.DataLayer do
              upsert_fields: upsert_fields,
              return_records?: true
            }) do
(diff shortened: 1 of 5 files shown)


zachdaniel (approved):

zachdaniel:

PR #381, opened 2024-09-06
chore: failing test for constraint violation with stream strategy

rbino:
This is probably a problem in Ash but I didn't know a simple way to make the ETS data layer fail similarly to a constraint violation

### Contributor checklist


zachdaniel:

zachdaniel:
Fixed in `main` of `ash`.

PR #385, opened 2024-09-11
Implement Ltree Type

maennchen:
### Contributor checklist

Introduces new `AshPostgres.Ltree` type.

The type contains an `:escape?` constraint that allows to put any UTF-8 value into the ltree path segments. I'm using the escape functionality at the moment to put UUIDs (`-` not allowed in Postgres 15) and strings which contain a dot into the path segments.

As a follow up, `lquery` and various operators / function could also be implemented.


zachdaniel (approved):

zachdaniel:
Awesome 🔥

zachdaniel:

PR #393, opened 2024-09-22
fix(installer): use correct module name in the `DataCase` moduledocs.

jimsynz:


zachdaniel:

PR #406, opened 2024-10-17
fix: race condition compiling migrations when concurrently creating new tenants

skrioify:
I ran into an issue where concurrently creating new tenants in async tests crashed due to ash_postgres attempting to concurrently compile the same tenant migration file.

### Contributor checklist

- [ ] Bug fixes include regression tests (attempted, but hard to ensure it wouldn't be flaky)

### Notes

I attempted to write a regression test as well, but came to the conclusion that it would end up being brittle since it is timing-based and needs multiple DB connections. It would also require me to add a whole new domain with context-based multitenancy instead of attribute-based to the test repo.

### DIFF
--- a/lib/migration_compile_cache.ex
+++ b/lib/migration_compile_cache.ex
@@ -1,0 +1,38 @@
+defmodule AshPostgres.MigrationCompileCache do
+  @moduledoc """
+  A cache for the compiled migrations.
+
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:

PR #418, opened 2024-11-06
Fix Complex aggregates doesn't honor tenant

pshoukry:
Following the steps [here](https://elixirforum.com/t/using-an-ash-resource-with-a-dynamic-repository/61138) to use multi-tenancy with dynamic repo. when complex aggregates are called:
```
aggregates do
    sum :saved_money, [:redeems, :deal], :amount do
  end
```

When Using Ash.load with tenant, I can see the tenant used in the Tracer switching:

The first call is done with the tenant passed but 

the second sum aggregation query is selecting with the default tenant returning no associated records hence sum is 0.

The goal of the PR is to reproduce the issues using tests. The tests need to dynamically create and migrate the 2 tenant dbs and check on the data.
Fix the issue 

### Contributor checklist


zachdaniel:
Hello! Could you please add some context on this PR?

pshoukry:
@zachdaniel I updated the context and will continue working on it .. my apologies for not closing sooner but i got swamped.

test/aggregate_test.exs
```diff
@@ -57,6 +57,40 @@ defmodule AshSql.AggregateTest do
...
+      |> Enum.map(
+        &(Post
+          |> Ash.Changeset.for_create(:create, %{name: &1, user_id: user.id})
+          |> Ash.create!(tenant: "org_#{org.id}", load: [:last_word]))
+      )
+
+      assert Ash.load!(user, :count_visited, tenant: "org_#{org.id}")
+             |> then(& &1.count_visited) == 4
+
+      assert Ash.load!(org, :total_posts)
```
pshoukry, line 86: @zachdaniel I think I was able to get it working ( sorry was in sort of a rabbit hole) .. 

1. Does this scenario supposed to be supported to start with ? It works for me without multi-tenancy
2. It for now fails cause it can't find the schema 

Am I doing something wrong or deviating for original blue print or should i try to make it work ?

test/aggregate_test.exs
```diff
@@ -57,6 +57,40 @@ defmodule AshSql.AggregateTest do
...
+             |> then(& &1.total_posts) == 4
+
+      assert Ash.load!(org, :total_users_posts, tenant: "org_#{org.id}")
```
pshoukry, line 89: Fails with and without explicit tenant

pshoukry:
@zachdaniel I got my tests working ( and failing ) can you take a look at my comments and let me know If I am missing anything ? If you are ok I will try to work on the fix.

test/aggregate_test.exs
zachdaniel: Yes, this should work as your test shows. Great work!

zachdaniel:
Looks correct to me!

pshoukry:
@zachdaniel will pull latest and will try to figure it out .. thank you for all the effort you put in ash ;)

PR #434, opened 2024-12-02
Complex aggregates doesn't honor tenant

pshoukry:
Following the steps [here](https://elixirforum.com/t/using-an-ash-resource-with-a-dynamic-repository/61138) to use multi-tenancy with dynamic repo. when complex aggregates are called:
```
aggregates do
    sum :saved_money, [:redeems, :deal], :amount do
  end
```

When Using Ash.load with tenant, I can see the tenant used in the Tracer switching:

The first call is done with the tenant passed but 

the second sum aggregation query is selecting with the default tenant returning no associated records hence sum is 0.


zachdaniel:
@pshoukry so these tests illustrate an existing issue? We did fix some aggregate query building issues, would be good to confirm if this issue is still present in the latest version.  if so we can merge and. I can investigate.

pshoukry:
@zachdaniel I already merged main and the issue is still happening .. I tried to track the issue but it goes through many layers and I can't dedicate enough time.. you would definitely know your way better than me .. so agree would be great if you can merge and take a look

zachdaniel:
@pshoukry I merged this a bit hastily I guess. I've had to remove the tests from the main branch. They don't appear to be calling real loads. Can you adjust and resubmit? They are failing due to "no such load".

pshoukry:
I am really sorry will open a new one with the 2 loads, sorry i was playing with them locally trying to figure out the issue.

PR #435, opened 2024-12-04
docs: add configuration example and reference in exceptions

binarypaladin:
This stemmed from questions in Discord about `:where` in conjunction with identities. The messages are clearer now that I get it. The problem is, when you don't get it, the dot notation is like, wat.

I beefed up the docs for `AshPostgres.DataLayer.Info.identity_wheres_to_sql/1` and then referenced them from everywhere else that I could find that made sense.

The reason for this is because of my own process prior to just asking.

1. I got the exception and looked at it.
2. I went to the docs for this package and searched for `identity_wheres_to_sql`.
3. I got the original docs: "A keyword list of identity names to the sql representation of their where clauses"

While said documentation is factually correct... without context I was confused.

Now, the error leads to a place that can be easily referenced. (I didn't want to constantly copy examples.) This follows my typical pattern for documentation. 

Also, I added specs. I know there are none there, and perhaps I should avoid this in the future, but I really like specs in documentation in most cases and this one was simple enough.

Obviously, if this doesn't conform to standards, lemme know and I'll adjust.

### Contributor checklist


lib/data_layer.ex
```diff
@@ -2657,9 +2657,11 @@ defmodule AshPostgres.DataLayer do
       case identity do
         %{name: name, where: where} when not is_nil(where) ->
           AshPostgres.DataLayer.Info.identity_where_to_sql(resource, name) ||
-            raise(
-              "Must provide an entry for :#{identity.name} in `postgres.identity_wheres_to_sql` to use it as an upsert_identity"
-            )
+            raise("""
+            Must provide an entry for :#{identity.name} in `postgres.identity_wheres_to_sql` to use it as an upsert_identity.
+
+            See `AshPostgres.DataLayer.Info.identity_wheres_to_sql/1` for an example.
```
zachdaniel, line 2663: exception messages w/ backticks like this is a bit annoying because it doesn't resolve to an actual URL. I'd suggest linking directly to hex so users have something to click on.

zachdaniel (approved):

zachdaniel:
Had one piece of feedback but otherwise LGTM. One note: if you ever see anything without type specs or docs and you'd like to add them, please do not hesitate. Every public function should have specs and docs ❤️

binarypaladin:
> Every public function should have specs and docs eventually ❤️

Cool. Will do. It might do it randomly as I mess with things. Sometimes I go on a tear in a single module.

Anyway, changes made.

zachdaniel:

PR #438, opened 2024-12-05
improvement: add `postgres_reference_expr` callback to AshPostgres.Type

stevebrambilla:
We have a custom Ash type to help map integer columns in an existing database schema to booleans in Elixir. 

Ash.Type gets us most of the way there, but to support bare attribute filter expressions like `active?` we need to generate a boolean expression for the WHERE clause. This adds another callback to AshPostgres.Type so the implementation can rewrite those expressions when generating the query, eg. with a `? != 0` fragment for our integer-boolean example.

### Contributor checklist

### DIFF
--- a/lib/sql_implementation.ex
+++ b/lib/sql_implementation.ex
@@ -154,6 +154,33 @@ defmodule AshPostgres.SqlImplementation do
   end
 
   def expr(
+    query,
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:

PR #443, opened 2024-12-13
fix bug involving duplicate foreign key constraint names across schema

olivermt:
alter resource generation query to go to the source pg_constraints table instead of to the view to fetch constraint data

This path is already tested, I sadly don't have time to map up duplicate names on references to test now, should maybe add issue to track that chore?


olivermt:
I closed due to a failing test on resource generation, but it was failing before my commit so I am reopening :P

zachdaniel:

PR #444, opened 2024-12-15
Fix query for metadata on foreign keys and fix duplicate references being produced

olivermt:
Title says it all.


zachdaniel:

PR #456, opened 2025-01-08
Failing test for module calculation inside expression one

vonagam:


test/comedy_test.exs
```diff
@@ -0,0 +1,11 @@
+defmodule AshPostgres.Test.ComedyTest do
```
zachdaniel, line 1: can we put this into the calculations test module?

vonagam: Done.

PR #457, opened 2025-01-10
test: add test for relationship filter with `parent(parent(field))`

vonagam:
The test currently fails.

`parent(parent(id))` behaves as `parent(id)`.

### DIFF
--- a/test/rel_with_parent_filter_test.exs
+++ b/test/rel_with_parent_filter_test.exs
@@ -1,7 +1,7 @@
 defmodule AshPostgres.RelWithParentFilterTest do
   use AshPostgres.RepoCase, async: false
 
-  alias AshPostgres.Test.Author
+  alias AshPostgres.Test.{Author, Comedian, Joke}
 
   require Ash.Query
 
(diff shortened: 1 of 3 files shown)


zachdaniel:
The fix for this is unfortunately complex. What *should* be done to resolve this in the general case is to not use simple integers for the binding names, but to use something like `{identifier, num}`. actually typing that out it might not be *too* hard, but it will take enough time that I don't think I can do it in the near future.

Please open an issue to track this work. If you or someone else would like to do it, the general pattern would be to:

1. generate a unique id for each query we generate. We can use something like `System.monotonic_integer`, or even `make_ref`. That parts not important really.
2. when we bind something, we add the query's `id` to the binding, i.e `{id, binding}`
3. Whenever we do a lateral join, we ensure that the query we are lateral joining to has its own identifier

zachdaniel:
Please remind me when you open the issue to add it to the roadmap under the "soon" category.

zachdaniel:
In the meantime, you can potentially accomplish what you need to accomplish with manual relationships or custom calculations.

vonagam:
Ok, opened the issue.

Yeah, I was using `fragment` workaround but thought about converting it into argumented calculation.

(Reminder about adding "soon" category to the issue.)

PR #464, opened 2025-01-16
feat: add options to disable atomic actions and error expressions

barnabasJ:
### Contributor checklist


zachdaniel:
Actually @barnabasJ we should make this a repo callback, not compile time config.

Issue #462, opened 2025-01-15
Possible unsafe migrations warning


dkuku:
I'm working through the book and I see that when a migration is generated for new index it's not applying `concurrently: true`
```
  def up do
    create unique_index(:albums, [:name, :artist_id],
             name: "albums_unique_album_names_per_artist_index"
           )
  end
```
This locks the table and with bigger databases can cause problems:
[postgres docs](https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY)
there are some libraries that will let you know about unsafe migrations where you can check what to avoid
[rails](https://github.com/ankane/strong_migrations?tab=readme-ov-file#adding-an-index-non-concurrently) - there is also one for [elixir](https://hexdocs.pm/excellent_migrations/readme.html) but as ash generates the migrations most of the stuff can be applied by changing the codegen.

zachdaniel:
There are some extra considerations with concurrent index creation, because it can't be done in a transaction. We have tools for it with custom indexes and we could potentially do it by default or allow the default to be configured for identities.

zachdaniel:
Yeah, I think we could do it by default for identities on tables that we didn't just create, and shuffle them into their own separate migration with the transaction disabled like we do for concurrent custom indexes. PRs welcome!

PR #471, opened 2025-01-31
Improvement: generate migrations task support concurrent indexes flag

ken-kost:
#462 

### Contributor checklist


lib/mix/tasks/ash_postgres.generate_migrations.ex
```diff
@@ -22,6 +22,7 @@ defmodule Mix.Tasks.AshPostgres.GenerateMigrations do
   * `dry-run` - no files are created, instead the new migration is printed
   * `check` - no files are created, returns an exit(1) code if the current snapshots and resources don't fit
   * `snapshots-only` - no migrations are generated, only snapshots are stored
+  * `concurrent-indexes` - new identities will be run in a separate migration (like concurrent custom indexes)
```
zachdaniel, line 25:
```suggestion
  * `concurrent-indexes` - new identities will be run concurrently and in a separate migration (like concurrent custom indexes)
```

PR #491, opened 2025-02-26
Peter/fix schema multitenant aggregates wrong tenant

pshoukry:
### Contributor checklist


pshoukry:
@zachdaniel sorry this is the test with the 2 loads. It fails cause its trying to load from the wrong database.

zachdaniel:
So the issue in this case was that the `tenant` option was not being set, but you should have gotten a better error. Additionally, we should have been able to infer the tenant from the record. I'm adding this behavior to ash core.

pshoukry:
@zachdaniel thank you for looking in to it

pshoukry:
@zachdaniel sounds great will appreciate if you can mention me in the related PR so I can understand the issue more

zachdaniel:
Its on this commit in `ash` `7909909a16af47a22a7e3cdbcd1a8244cc4ef499`

Issue #490, opened 2025-02-25
Add mix task to undo latest codegen/delete latest migration


sevenseacat:
**Is your feature request related to a problem? Please describe.**

Codegen to generate a migration is really awesome, but sometimes you generate a migration and then you realize you missed something, so you want to delete that migration, make the change, and then generate it again. Currently that's a manual process - go find the migration, find the snapshots for all the relevant tables, delete them all, and do it again. It's pretty annoying!

**Describe the solution you'd like**

A task like `ash_postgres.delete_latest_migration` that will delete the last migration, and any snapshots that were created at the same time. No option to delete other migrations or anything like that - only the last one.

And it should probably confirm the name of the migration just to make sure!

**Describe alternatives you've considered**

You could just generate another migration with the added changes, but snapshot files are big and ugly and if you haven't even run the first migration, being able to undo/regenerate it seems like the smarter approach.

zachdaniel:
There is an example here of a script that does it based off of git status:

https://hexdocs.pm/ash_postgres/migrations-and-tasks.html#regenerating-migrations

zachdaniel:
I don't think we can just do "only the last one" because it is actually possible for the migration generator to produce *more than one* migration (concurrent index generation must be done in its own migration w/o a ddl transaction).

sevenseacat:
Oh interesting! Yeah if I don't feel like doing it manually I usually do it with git now - add, stash, drop - but that feels a bit weird.

Okay so maybe not the *last* one, but any with the timestamp of the last one? If more than one migration is generated, do they have the same timestamp?

zachdaniel:
They do not, the timestamp differentiates them. However, I think the fix here is actually to support a `--dev` flag for migrations, which uses numbered migrations like `Dev1` and `Dev2` etc. Then, when migrations are checked with `mix ash.codegen --check`, we error if there are dev migrations. When you run `mix ash.codegen name` we rollback all `dev` migrations, delete all dev snapshots and regenerate them fully.

The added benefit here (which is a big one) is we could then get:
1. a watcher that auto generates migrations with `--dev` as you change your files 
2. a plug like Ecto's plug that says "you haven't run codegen yet, click here to run it"

zachdaniel:
and then `mix ash.codegen --reset-dev` or something would then do what you want, based on dev migrations, not named migrations. Then you just name it once at the end when you're done with your feature.

zachdaniel:
[...]

PR #496, opened 2025-03-02
feat: dev flag for generating migrations

ken-kost:
#490 

> I think the fix here is actually to support a --dev flag for migrations, which uses numbered migrations like Dev1 and Dev2 etc. Then, when migrations are checked with mix ash.codegen --check, we error if there are dev migrations. When you run mix ash.codegen name we rollback all dev migrations, delete all dev snapshots and regenerate them fully.

For now I'm just adding `_dev` everywhere and I left the timestamp which gives it uniqueness.
For now just deleting `_dev` files when `--dev` is not provided. 

Not sure if in the right direction, need feedback. Putting this into draft.
### Contributor checklist

### DIFF
--- a/test/dev_migrations_test.exs
+++ b/test/dev_migrations_test.exs
@@ -125,12 +125,88 @@ defmodule AshPostgres.DevMigrationsTest do
     end
   end
 
+  describe "--dev option tenant" do
+    setup do
+      on_exit(fn ->
+        resource_dev_path = "priv/resource_snapshots/dev_test_repo"
+        resource_files = File.ls!(resource_dev_path)
+        Enum.each(resource_files, &File.rm_rf!(Path.join(resource_dev_path, &1)))
+        migrations_dev_path = "priv/dev_test_repo/migrations"
+        migration_files = File.ls!(migrations_dev_path)
+        Enum.each(migration_files, &File.rm!(Path.join(migrations_dev_path, &1)))
+        tenant_migrations_dev_path = "priv/dev_test_repo/tenant_migrations"
+        tenant_migration_files = File.ls!(tenant_migrations_dev_path)
+        Enum.each(tenant_migration_files, &File.rm!(Path.join(tenant_migrations_dev_path, &1)))
+      end)
+    end
+
+    test "rolls back dev migrations before deleting" do
+      defposts do
+        postgres do
+          schema("example")
+        end
+
+        attributes do
+          uuid_primary_key(:id)
+          attribute(:title, :string, public?: true, primary_key?: true, allow_nil?: false)
+        end
+
+        multitenancy do
+          strategy(:context)
+        end
+      end
+
+      defdomain([Post])
+      capture_log(fn -> tenant_migrate() end) |> dbg()
+
+      AshPostgres.MigrationGenerator.generate(Domain,
+        snapshot_path: "priv/resource_snapshots",
+        migration_path: "priv/dev_test_repo/migrations",
+        tenant_migration_path: "priv/dev_test_repo/tenant_migrations",
+        dev: true
+      )
+
+      assert [] =
+               Enum.sort(
+                 Path.wildcard("priv/dev_test_repo/migrations/**/*_migrate_resources*.exs")
+               )
+               |> Enum.reject(&String.contains?(&1, "extensions"))
+
+      assert [_tenant_migration] =
+               Enum.sort(
+                 Path.wildcard("priv/dev_test_repo/tenant_migrations/**/*_migrate_resources*.exs")
+               )
+               |> Enum.reject(&String.contains?(&1, "extensions"))
+
+      assert capture_log(fn -> tenant_migrate() end) =~ "create table posts"
+      assert capture_log(fn -> tenant_migrate() end) =~ "create table posts"
+
+      AshPostgres.MigrationGenerator.generate(Domain,
+        snapshot_path: "priv/resource_snapshots",
+        migration_path: "priv/dev_test_repo/migrations",
+        tenant_migration_path: "priv/dev_test_repo/tenant_migrations"
+      )
+
+      assert capture_log(fn -> tenant_migrate() end) =~ "create table posts"
+    end
+  end
+
   defp migrate do
     Mix.Tasks.AshPostgres.Migrate.run([
       "--migrations-path",
       "priv/dev_test_repo/migrations",
       "--repo",
       "AshPostgres.DevTestRepo"
+    ])
+  end
+
+  defp tenant_migrate do
+    Mix.Tasks.AshPostgres.Migrate.run([
+      "--migrations-path",
+      "priv/dev_test_repo/tenant_migrations",
+      "--repo",
+      "AshPostgres.DevTestRepo",
+      "--tenants"
     ])
   end
 end


ken-kost:
To try out make some changes and run
`mix ash_postgres.generate_migrations --dev`
it will create snapshots and migrations with `_dev` ending.
`mix ash_postgres.generate_migrations`
will remove all migrations and snapshots with `_dev` ending and continue as usual.

zachdaniel:
This is great! I've got a busy week ahead of me, but will take a look when I get the chance 🙇

test/migration_generator_test.exs
```diff
@@ -2345,4 +2333,11 @@ defmodule AshPostgres.MigrationGeneratorTest do
                ~S[modify :post_id, references(:posts, column: :id, name: "comments_post_id_fkey", type: :uuid, prefix: "public")]
     end
   end
+
+  defp get_migration_files(opts) do
```
ken-kost, line 2337: @zachdaniel What do you think of this helper function and default opts in setup? I like this because it adds clarity in test. I could refactor the whole file if welcomed. 🐛

zachdaniel: I like it 👍 lets refactor that in a different PR though.

zachdaniel:
So @sevenseacat has made two good points about the current solution here:

1. we need to rollback these migrations in the dev & test databases before deleting them. I *think* (will need some experimentation) that we should do that with `System.cmd("mix", ["ecto.rollback", "--to", "<the earliest migration id>])`, and then again with `MIX_ENV=test` in the environment. We will also need to roll it back for any tenants if there are tenant migrations. The logic for that is a bit hairier but basically consists of, for each env, listing all tenants and and running the same command with `--prefix <the tenant>`.

2. we should add a variation of this task that works using git history. What it would do is mark all uncommitted migration files as dev migrations, and then go through this same path. Something like `mix ash.codegen --squash-uncommitted`

I think #1 will need to be a part of this PR because otherwise we'll be putting users dbs in a bad state (they can't roll back the migrations because they've been deleted).

#2 can be done in a followup.

zachdaniel:
Do the tests cover the rollback behavior?

ken-kost:
> Do the tests cover the rollback behavior?

right, I'll expand it calling migrate. 👍

ken-kost:
@zachdaniel 
Okay so I'm having some trouble doing rgw migration in test, my idea was to assert it was made successfully both times after generating migrations, dev and regular. 
What happens is that twice the same extension migration is made (for sandbox repo and for test repo) which breaks the migration. For instance if I do
`Mix.Tasks.AshPostgres.Migrate.run(["--migrations-path", "test_migration_path"])`
I would get
```
** (Ecto.MigrationError) migrations can't be executed, migration name migrate_resources_extensions_1 is duplicated
```
Also haven't found any test running migrate or rollback 🤔 
Do you have an idea how should I solve this? I'm thinking somehow getting rid of the sandbox repo maybe, I'm not sure. 🐛

zachdaniel:
Hmm...I think we should be making the sandbox and non sandbox repo use a different path for their migrations then?

ken-kost:
> Hmm...I think we should be making the sandbox and non sandbox repo use a different path for their migrations then?

Okay so it gets tricky. Because migration path for any migration is determined:
```
  defp migration_path(opts, repo, tenant? \\ false) do
    # Copied from ecto's mix task, thanks Ecto ❤️
    config = repo.config()

    if tenant? do
      if path = opts.tenant_migration_path || config[:tenant_migrations_path] do
        path
      else
        priv =
          AshPostgres.Mix.Helpers.source_repo_priv(repo)

        Path.join(priv, "tenant_migrations")
      end
    else
      if path = opts.migration_path || config[:tenant_migrations_path] do
        path
      else
        priv =
          AshPostgres.Mix.Helpers.source_repo_priv(repo)

        Path.join(priv, "migrations")
      end
    end
  end
```
So it's either configured through `opts` or is `AshPostgres.Mix.Helpers.source_repo_priv(repo)`. So if `migration_path` is set then both repo's hit that path. If not then it hits existing migrations, but we are creating here temporary migrations in a temporary folder right, also, using existing macros like `defposts` I think it triggers in migration a `CREATE TABLE posts` which already exists in existing migrations (I did circumvent this by adding a macro that creates a new table but I still had issues) . 

This simple assertion turned out way harder than anticipated. 😓

ken-kost:
I'll work on it more, try to find a path. 🐛

lib/migration_generator/migration_generator.ex
```diff
@@ -491,6 +509,55 @@ defmodule AshPostgres.MigrationGenerator do
...
+    |> case do
+      {:error, _error} -> []
+      {:ok, migrations} -> Enum.filter(migrations, &String.contains?(&1, "_dev.exs"))
+    end
+  end
+
+  defp remove_dev_migrations(dev_migrations, tenant?, repo, opts) do
+    version = dev_migrations |> Enum.min() |> String.split("_") |> hd()
+    test_env = [env: [{"MIX_ENV", "test"}]]
+    System.cmd("mix", ["ash_postgres.rollback", "--to", version])
```
zachdaniel, line 525: This should use:

```elixir
Mix.Task.reenable("ash_postgres.rollback")
Mix.Task.run("ash_postgres.rollback", ["--repo", inspect(repo)])
```

(using your new `--repo` option). This also avoids needing to do anything special with `MIX_ENV` etc.

lib/mix/tasks/ash_postgres.generate_migrations.ex
```diff
@@ -21,6 +21,7 @@ defmodule Mix.Tasks.AshPostgres.GenerateMigrations do
   * `no-format` - files that are created will not be formatted with the code formatter
   * `dry-run` - no files are created, instead the new migration is printed
   * `check` - no files are created, returns an exit(1) code if the current snapshots and resources don't fit
+  * 'dev' - dev files are created
```
zachdaniel, line 24:
```suggestion
  * `dev` - dev files are created
```

zachdaniel:
Added some feedback. We can get this over the line and I'll merge and figure out what to do about testing the rollback 😄

ken-kost:
Now I'm thinking maybe the `-r` flag could also help with the testing. 🤔 I'll check it out but if so, what a fortunate coincidence 🍀

lib/migration_generator/migration_generator.ex
ken-kost: After doing this, if I leave
```
    if tenant? do
      Enum.each(repo.all_tenants(), fn tenant ->
        opts = ["--repo", inspect(repo), "--to", version, "--prefix", tenant]
        Mix.Task.run("ash_postgres.rollback", opts)
      end)
    end
```
then it fails. I get an error
```
     ** (Ash.Error.Unknown) 
     Bread Crumbs:
       > Error returned from: AshPostgres.MultitenancyTest.Org.read
```
which comes from `repo.all_tenants()`
and there is also a log
```
     The following output was logged:
     11:30:35.118 [info] Migrations already down
     11:30:35.123 [info] Migrations already down
```

So I assume that also being done.? 🤔 I removed it then.

ken-kost:
Okay so 
```
      Mix.Tasks.AshPostgres.Migrate.run([
        "--migrations-path",
        "test_migration_path",
        "--repo",
        "AshPostgres.TestRepo"
      ])
```
worked, but only once. 😓 i.e. it works after `test.reset` and does not work the second time and also does not work when I run it again after the real generate migration.

And it worked if I commented out `create_extension_migrations(repos, opts)` and in test made async true ` use AshPostgres.RepoCase, async: true` FYI

zachdaniel:
I think you need to re-migrate as part of test teardown.

zachdaniel:
@ken-kost making some relatively substantial changes, opening a new PR (still based off of your work)

PR #497, opened 2025-03-03
fix: handle errors from identities in polymorphic resources properly

vonagam:
Was getting `Ash.Error.Unknown` from identity constraints in polymorphic resources on violations because table context information was discarded at that place in code and `Ecto.Changeset.unique_constraint` was getting called with wrong constraint names.


zachdaniel:

PR #505, opened 2025-03-09
enhancement: migration generator test refactor

ken-kost:
Okay so this is what I had in mind but now that I see it it ain't much. 😅 In the sense that such a refactor was uncalled for and it didn't improve much. I like it this way more but that's just my opinion, the tests were readable enough without this. It's a lot of grunt work, which I don't mind, I could refactor tests all day, but little is gained I think so... Thoughts?
### Contributor checklist


ken-kost:
For some reason I can't point this to my other PR, probably because of the fork. 🤔

PR #515, opened 2025-03-24
fix parsing of constraints and indexes that are wrapped in quotes

olivermt:


zachdaniel:

PR #521, opened 2025-03-29
improvement: propagate `-r` flag to Ecto

ken-kost:
### Contributor checklist

### DIFF
--- a/lib/mix/tasks/ash_postgres.rollback.ex
+++ b/lib/mix/tasks/ash_postgres.rollback.ex
@@ -59,9 +59,10 @@ defmodule Mix.Tasks.AshPostgres.Rollback do
           log_migrations_sql: :boolean,
           log_migrator_sql: :boolean,
           only_tenants: :string,
-          except_tenants: :string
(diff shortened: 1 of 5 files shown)


lib/mix/tasks/ash_postgres.rollback.ex
```diff
@@ -59,9 +59,10 @@ defmodule Mix.Tasks.AshPostgres.Rollback do
...
           only_tenants: :string,
-          except_tenants: :string
+          except_tenants: :string,
+          repo: :string
         ],
-        aliases: [n: :step, v: :to]
+        aliases: [n: :step, v: :to, r: :repo]
       )
 
     repos = AshPostgres.Mix.Helpers.repos!(opts, args)
```
zachdaniel, line 68: `AshPostgres.Mix.Helpers` should be updated to detect the `repos` option, and should return those repos only.

zachdaniel:

zachdaniel:
Ah, I merged this, but we also need to document these in the task docs, can you follow up with that?

Issue #459, opened 2025-01-13
Inconsistency between documentation and actual `ash_postgres_subquery` type


lusergit:
When talking about manual relationships in the [HexDocs](https://hexdocs.pm/ash_postgres/manual-relationships.html) and in the [behavior reference manual](https://hexdocs.pm/ash_postgres/manual-relationships.html) the return type of `ash_postgres_subquery` is said to be of the type `{:ok, Ecto.Query.t()} | {:error, error}`, when in reality when used in the library its implicit type is just `Ecto.Query.t()` 

https://github.com/ash-project/ash_postgres/blob/8e59757864ce65c6581f6fdb779332291d5c06b4/lib/data_layer.ex#L1032-L1058

zachdaniel:
Ah, right you are. Would you like to Pr a correction?

PR #524, opened 2025-04-08
fix: ash postgres subquery usage

artiom:
Closes https://github.com/ash-project/ash_postgres/issues/459

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -1041,12 +1041,16 @@ defmodule AshPostgres.DataLayer do
         Map.get(relationship, :manual) ->
           {module, opts} = relationship.manual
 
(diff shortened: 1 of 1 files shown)


artiom:
Couldn't find a way how to test that code path. `manual` relation has unique paths (in `ash` lib), couldn't figure out how to get to this code snippet with tests. Happy for any feedback.

zachdaniel:
I think for backwards compatibility we need to accept both a query return or an ok/error tuple return.

artiom:
> I think for backwards compatibility we need to accept both a query return or an ok/error tuple return.

Added.

zachdaniel (approved):

PR #536, opened 2025-04-23
fix: use old multitenancy in generated removals of previous indexes

vonagam:
A previous index was created with previous tenancy and so when it is being removed it should use it too, not current one.

### DIFF
--- a/lib/migration_generator/operation.ex
+++ b/lib/migration_generator/operation.ex
@@ -1219,29 +1219,38 @@ defmodule AshPostgres.MigrationGenerator.Operation do
 
   defmodule RemoveCustomIndex do
     @moduledoc false
-    defstruct [:schema, :table, :index, :base_filter, :multitenancy, no_phase: true]
+    defstruct [
+      :schema,
+      :table,
+      :index,
+      :base_filter,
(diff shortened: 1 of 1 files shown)


lib/migration_generator/operation.ex
```diff
@@ -1219,29 +1219,29 @@ defmodule AshPostgres.MigrationGenerator.Operation do
 
   defmodule RemoveCustomIndex do
     @moduledoc false
-    defstruct [:schema, :table, :index, :base_filter, :multitenancy, no_phase: true]
+    defstruct [:schema, :table, :index, :base_filter, :multitenancy, :old_multitenancy, no_phase: true]
```
zachdaniel, line 1222: Do we need to update where we create this and add `old_multitenancy`?

vonagam: Works like that because of [this line](https://github.com/ash-project/ash_postgres/blob/0124cb4c82f9327c3565103f0a3f135f1292a04f/lib/migration_generator/migration_generator.ex#L2113) that adds `old_multitenancy` field to all operations.

zachdaniel:

Issue #540, opened 2025-05-01
Error in Recursive Relationships example


user20230119:
I get `* ** (BadMapError) expected a map, got: [...]` on this line of the Recursive Relationships example.

https://github.com/ash-project/ash_postgres/blame/097509c7433afa735d2326cbed1388cce2803cc0/documentation/topics/advanced/manual-relationships.md#L136

PR #542, opened 2025-05-02
edit manual relationships example with group_by and relationship name

user20230119:
The previous commit for https://github.com/ash-project/ash_postgres/issues/540 set `employee.descendants` as `[]`.

I think you meant to do `Enum.group_by`.


zachdaniel:
This does look better, thanks. I'm busy and NGL was kind of winging that last fix. Much appreciated.

PR #576, opened 2025-06-26
fix: escape " in check constraint generator

frankdugan3:
# Contributor checklist

- [x] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/migration_generator/operation.ex
+++ b/lib/migration_generator/operation.ex
@@ -1351,6 +1351,8 @@ defmodule AshPostgres.MigrationGenerator.Operation do
           },
           table: table
         }) do
+      check = String.replace(check, ~S["], ~S[\"])
+
       if base_filter do
         "create constraint(:#{as_atom(table)}, :#{as_atom(name)}, #{join(["check: \"(#{check}) OR NOT (#{base_filter})\")", option(:prefix, schema)])}"
       else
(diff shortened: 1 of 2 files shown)


frankdugan3:
Actually, I think there's still some problems. Let me investigate further before you check on this.

frankdugan3:
OK, it's good now, derp! :zany_face:

zachdaniel (approved):

zachdaniel:
Actually, we have to handle already escaped double quotes only right? otherwise this effectively unescapes them, turning `\"` into `\\"`. So we ned to replace all unescaped double quotes i.e `"` that start w/ zero or an od number of `\`.

zachdaniel:
Thinking about this more, I don't think we need to "fix" this? If you want double quotes in that code, you just need to do `\\"`. closing for now, but feel free to discuss further if you disagree.

frankdugan3:
Well, it's up to you but I do somewhat strongly disagree because it's inconsistent syntax compared to custom statements and fragments. And if you use multiline sigils, it's very strange, and if using something like `~SQL`, it breaks syntax highlighting.

I agree what I did is an ugly solution, though.

Custom statements do this, which bypasses the problem entirely:

```elixir
def up(%{statement: %{up: up, code?: false}}) do
  """
  execute(\"\"\"
  #{String.trim(up)}
  \"\"\")
  """
end
```

Perhaps a simpler refactor to do this instead would be a better middle ground like this:

```elixir
"""
create constraint(:#{as_atom(table)}, :#{as_atom(name)}, check: \"\"\"
#{String.trim(check)}#{option(:prefix, schema)}
\"\"\"
"""
```

That would be consistent with the other generators, and sidestep my original mess.

PR #579, opened 2025-06-30
raise error if the constraint is created on a field that doesn't exist

xantrac:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/migration_generator/migration_generator.ex
+++ b/lib/migration_generator/migration_generator.ex
@@ -3125,6 +3125,12 @@ defmodule AshPostgres.MigrationGenerator do
             resource
             |> Ash.Resource.Info.attribute(attribute)
 
+          if is_nil(attr) do
+            raise """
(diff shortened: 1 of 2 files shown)


zachdaniel:
Thanks for the PR! I think this behavior is valuable but would be better placed in a verifier on the data layer extension. This will give errors at compile time instead of when generating migrations. See this verifier for example: https://github.com/ash-project/ash_postgres/blob/main/lib/verifiers/validate_identity_index_names.ex

xantrac:
> Thanks for the PR! I think this behavior is valuable but would be better placed in a verifier on the data layer extension. This will give errors at compile time instead of when generating migrations. See this verifier for example: https://github.com/ash-project/ash_postgres/blob/main/lib/verifiers/validate_identity_index_names.ex

Sounds good!

PR #580, opened 2025-07-01
failing test for parent ref / policy thing

kernel-io:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/support/resources/post.ex
+++ b/test/support/resources/post.ex
@@ -611,7 +611,7 @@ defmodule AshPostgres.Test.Post do
     has_many(:posts_with_my_organization_name_as_a_title, __MODULE__) do
       public?(true)
       no_attributes?(true)
-      filter(expr(fragment("? = ?", organization.name, parent(organization.name))))
+      filter(expr(fragment("? = ?", title, parent(organization.name))))
     end
 
     belongs_to :parent_post, __MODULE__ do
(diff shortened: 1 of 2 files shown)


test/parent_filter_policy_test.ex
```diff
@@ -0,0 +1,35 @@
+defmodule ParentFilterPolicyTest do
+  use AshPostgres.RepoCase, async: false
+
+  alias AshPostgres.Test.{Organization, Post, User}
+
+  require Ash.Query
+
+  test "building references don't throw an exception when doing weird things" do
```
zachdaniel, line 8:
```suggestion
  test "when the first relationship in an `exists` path has parent references in its filter, we don't get errors" do
```

test/support/resources/post.ex
```diff
@@ -92,6 +92,12 @@ defmodule AshPostgres.Test.Post do
       authorize_if(relates_to_actor_via([:organization, :users]))
     end
 
+    policy action(:weird) do
```
zachdaniel, line 95: lets call this action something else, maybe `read_with_policy_with_parent`

zachdaniel:
@kernel-io this is fixed in latest `ash_sql` and `ash`, but can you address the feedback here so we can have the regression test?

 additionally, the test must be in a `.exs` file.

Issue #596, opened 2025-07-20
modify_query in read_action not applied in aggregation


TommasoAmici:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

ash 3.5.29
ash_postgres 2.6.11
elixir 1.18
erlang 27

### Operating system

macos

### Current Behavior

This is a sister bug to the one I opened in ash the other day https://github.com/ash-project/ash/issues/2195

I updated after the latest ash release but this is still a problem in ash_postgres as you anticipated.

### Reproduction

https://github.com/TommasoAmici/ash_postgres/commit/910ca508b4025c81aa64897736b47b0109a67610

### Expected Behavior

modify query is respected in action

PR #597, opened 2025-07-20
test: added regression test for aggregate with action with modify query

TommasoAmici:
Regression test for https://github.com/ash-project/ash_postgres/issues/596

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
Unfortunately, the fix here is going to be that we can't support aggregates that use read actions that have a `modify_query` option. Its just too complex to implement due to the various ways that aggregate queries are built.

zachdaniel:
I will add an error message explaining that this is not currently possible in Ash core.

Issue #582, opened 2025-07-04
Wrong order for the down/0 migration when changing both a field's name and nullability


ibarakaiev:
### Versions

Elixir: 1.18.4
OPT: 27

### Operating system

MacOS

### Current Behavior

I was introducing `AshStateMachine` to a resource which had an existing optional `status` field, which auto-generated the following migration:

```elixir
defmodule MyApp.Repo.Migrations.UpdateProducts do
  @moduledoc """
  Updates resources based on their most recent snapshots.

  This file was autogenerated with `mix ash_postgres.generate_migrations`
  """

  use Ecto.Migration

  def up do
    rename table(:products), :status, to: :state

    alter table(:products) do
      modify :state, :text, null: false
    end
  end

  def down do
    alter table(:products) do
      modify :status, :text, null: true
    end

    rename table(:product), :state, to: :status
  end
end
```

The `down/0` migration here fails because it attempts to modify `:status` before `:state` is renamed to `:status`.

### Reproduction

Have an existing resource with a nullable `status` field, then remove it, and introduce AshStateMachine (or just rename it to `:state` and make it nullable directly).

### Expected Behavior

The correct order should be:

```elixir
def down do
  alter table(:products) do
    modify :status, :text, null: true
  end

  rename table(:product), :state, to: :status
end
```

zachdaniel:
Thanks for the report! PRs welcome as I likely won't have time to look into this soon.

awerment:
I have a small, working fix. If nobody's picked this up yet, I would open a PR

PR #604, opened 2025-08-03
(fix) For renamed & modified attributes, use new attribute source in down migration

awerment:
fix(#582) For renamed & modified attributes, use new attribute source in down migration
    
When an attribute is renamed, alongside other changes to it (like changing default or not null constraint), the generated up migration first renames the column, then modifies it.
    
For the generated down migration, the order of operations is reversed, so we need to use the new column name when modifying the column, before renaming it back.

(i) There is one test failing for me locally, that I think is unrelated to this change (it's also failing on `main`)

### Contributor checklist

  - (modified existing unit test)
- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/migration_generator/operation.ex
+++ b/lib/migration_generator/operation.ex
@@ -752,7 +752,7 @@ defmodule AshPostgres.MigrationGenerator.Operation do
       up(%{
         op
         | old_attribute: op.new_attribute,
-          new_attribute: op.old_attribute,
+          new_attribute: Map.put(op.old_attribute, :source, op.new_attribute.source),
           old_multitenancy: op.multitenancy,
           multitenancy: op.old_multitenancy
       })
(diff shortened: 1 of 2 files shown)


lib/migration_generator/operation.ex
```diff
@@ -752,7 +752,7 @@ defmodule AshPostgres.MigrationGenerator.Operation do
       up(%{
         op
         | old_attribute: op.new_attribute,
-          new_attribute: op.old_attribute,
+          new_attribute: Map.put(op.old_attribute, :source, op.new_attribute.source),
```
zachdaniel, line 755: 🤔 I think this might have undesired consequences. This is one way to do it, but another way to do it would likely be to fix the operation sorting code. There is a `defp after?` function that we should likely make return `true` to make it so that any `modify` operations come after `rename` operations.

zachdaniel: Ah, right you are. To solve for that in the past we've split into separate operations that only run up and down. Looking at the code I don't think there are any issues actually.

zachdaniel:

PR #621, opened 2025-09-26
fix: use default constraint of 'now()' for AshPostgres.Timestamptz

siassaj:
NOTE: I haven't included any testing as it's unclear to me that there exist any tests for default constraint generators. I noticed that at least one previous patch applied to this region didn't include tests. LMK if I should do anything differently.

AshPostgres.Timestamptz creates a pg column of type 'timestamptz', however the existing default behaviour was to implement the default constraint as a 'timestamp' type. Effectively 'now()::timestamptz' was converted to 'timestamp' then back to 'timestamptz'. This has been changed to 'now()'.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #638, opened 2025-10-15
fix: handle results that can't be mapped to the changeset in bulk_create

### LINKED ISSUE #637: Ash.bulk_create/4 returns empty list records when using integer primary keys
### Code of Conduct

### AI Policy

### Versions

elixir 1.18.3
erlang 27.3
ash 3.6.2
ash_postgres 2.6.22

### Operating system

mac

### Current Behavior

Starting from ash_postgres 2.6.22, calling Ash.bulk_create/4 returns empty list records even when the insert actually succeeds in the database.

This regression seems related to the following commit:
https://github.com/ash-project/ash_postgres/commit/af0903f1380468bbe1b61fe65acb33772b5e24a5

From the code changes, it appears that the new identity-based result matching logic introduced in this commit fails when the primary key is integer-based (auto-increment), since the key is not present in the input changesets.

### Reproduction

_No response_

### Expected Behavior

_No response_

barnabasJ:
should fix #637 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -2123,32 +2123,8 @@ defmodule AshPostgres.DataLayer do
                 {Map.take(r, keys), r}
               end)
 
-            cant_map_results_to_changesets =
-              any_generated_keys_missing?(keys, resource, changesets)
-
             results =
-              if cant_map_results_to_changesets do
-                results
-                |> Enum.zip(changesets)
-                |> Enum.map(fn {result, changeset} ->
-                  if !opts[:upsert?] do
-                    maybe_create_tenant!(resource, result)
-                  end
-
-                  case get_bulk_operation_metadata(changeset) do
-                    {index, metadata_key} ->
-                      Ash.Resource.put_metadata(result, metadata_key, index)
-
-                    nil ->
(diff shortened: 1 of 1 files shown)


lib/data_layer.ex
```diff
@@ -2120,35 +2120,63 @@ defmodule AshPostgres.DataLayer do
...
               end)
 
-            results =
-              changesets
-              |> Enum.map(fn changeset ->
-                identity =
-                  changeset.attributes
-                  |> Map.take(keys)
+            cant_map_results_to_changesets =
+              any_generated_keys_missing?(keys, resource, changesets)
```
zachdaniel: but I don't see how this is possible? Don't we require the keys to be given to do an upsert on them? And this is only inside the upsert branch.

zachdaniel: Oh, then we should switch back to index based when it's not an upsert. If it's just a create, either they are all created or not, there is no variability. Can you make that change?

zachdaniel (approved):

PR #645, opened 2025-10-31
fix: added bulk_action_ref for changeset lookups

DGollings:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [X] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -1560,9 +1560,9 @@ defmodule AshPostgres.DataLayer do
                       :ok
                   end
 
-                  {:ok, [changeset.data]}
+                  {:ok, [add_bulk_metadata(changeset.data, changeset)]}
                 else
-                  {:ok, repo.all(query, repo_opts)}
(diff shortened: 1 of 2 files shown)


lib/data_layer.ex
```diff
@@ -1887,7 +1888,8 @@ defmodule AshPostgres.DataLayer do
             end)
 
           if options[:return_records?] do
-            {:ok, AshSql.Query.remap_mapped_fields(results, query)}
+            results = AshSql.Query.remap_mapped_fields(results, query)
+            {:ok, Enum.map(results, &add_bulk_metadata(&1, changeset))}
```
zachdaniel: For `update_query` and `destroy_query` this part doesn't really make much sense actually 🤔 So maybe it can really just be left out of here and can just be done in ash core when its iteratively doing the work. Also, don't we need to switch the bulk create logic to use refs now?

zachdaniel: Actually I guess it does make sense because update & destroy just call update_query/destroy... 🤔 I'm trying to determine if these are even strictly necessary. It may be only bulk creates that need to return their indexes? I will look into it 😄

PR #649, opened 2025-11-12
feat(aggregates): add multitenancy bypass for context

shahryarjb:
Based on https://github.com/ash-project/ash/pull/2427 for bypassing context aggregates

**Please for testing change the Ash to https://github.com/ash-project/ash/pull/2427  PR**; I did not change the mix.exs here so the tests are going to be failed

> Please review carefully; I might have missed something. Many parts were involved, but I tried to keep codebase changes minimal. All Ash tests passed successfully. It needs comprehensive checking the code, for example some place i could not be able to avg as sql, so i did in elixir

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -833,24 +833,29 @@ defmodule AshPostgres.DataLayer do
       repo = AshSql.dynamic_repo(resource, AshPostgres.SqlImplementation, query)
 
       with_savepoint(repo, query, fn ->
-        repo.all(
-          query,
-          AshSql.repo_opts(repo, AshPostgres.SqlImplementation, nil, nil, resource)
-        )
-        |> AshSql.Query.remap_mapped_fields(query)
-        |> then(fn results ->
-          if query.__ash_bindings__.context[:data_layer][:combination_of_queries?] do
-            Enum.map(results, fn result ->
-              struct(resource, result)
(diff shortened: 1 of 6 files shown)


lib/data_layer.ex
```diff
@@ -861,6 +866,146 @@ defmodule AshPostgres.DataLayer do
   defp no_table?(%{from: %{source: {"", _}}}), do: true
   defp no_table?(_), do: false
 
+  # Load bypass aggregates for each result record by querying across all tenants
+  defp load_bypass_aggregates(results, [], _resource, _original_query), do: {:ok, results}
```
zachdaniel: 🤔 I don't think we can realistically support this unfortunately. I think our only real option here is to forbid this behavior when using context multi tenancy and only allow it when using attribute multi tenancy. 

Looking at the code has highlighted to me that there will be a whole host of issues with this kind of approach.

For your needs, are you using attribute or context multitenancy? If it's just attribute multitenancy, it may be as straightforward as ensuring that when building the queries for aggregates not to add the tenant. That may even already work? 

If so, then I think the best bet is to forbid this behavior in Ash core, so no changes end up required here.

Apologies for suggesting this approach to you and then saying it can't be done 😢

zachdaniel:
Closing since we aren't planning on doing this now.

PR #654, opened 2025-11-18
test: multi relationship problems

barnabasJ:
This shows that there is a problem where both calculations loaded the same value because the read action was not checked when checking if the relationships are the same https://github.com/ash-project/ash/pull/2448

And it shows that the sort on the read action was not taking into account (makes the test flaky) https://github.com/ash-project/ash_sql/pull/194

# Contributor checklist

- [] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


barnabasJ:
I'll rebase once the other two PRs are released. Then I can also update the
dependencies in one go

zachdaniel:
just released it sorry.

PR #662, opened 2025-12-04
bugfix: Re-add `bulk_create_index` metadata when running bulk creates

sevenseacat:
This is necessary when running bulk create actions with the `sorted?: true` option, and also retains consistency with other built-in data layers

Related to https://github.com/ash-project/ash/issues/2452 and https://github.com/ash-project/ash_postgres/commit/4b6463b82f46938f0ac7b13af4c57fc52ff1f2b2

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -2142,8 +2142,12 @@ defmodule AshPostgres.DataLayer do
                     end
 
                     # Compatibility fallback
-                    Ash.Resource.put_metadata(
-                      result_for_changeset,
+                    result_for_changeset
+                    |> Ash.Resource.put_metadata(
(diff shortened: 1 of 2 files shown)


sevenseacat:
I have, and it works, but I wasn't entirely confident with your 'I haven't reproduced it but I tried something maybe it will work?' fix there, and the commit that removed `bulk_create_index` here looks like it was done accidentally.

If the official idea is that data layers don't set `bulk_create_index`, that's fine, I can fix all the other data layers to also do that (currently ash_postgres is the only one that doesn't).

PR #666, opened 2025-12-24
fix: apply join_relationship filter/sort/limit in many_to_many lateral joins

sevenseacat:
When a many_to_many relationship uses a join_relationship with sort/limit/filter, those constraints weren't being applied in the lateral join query.

Two issues:

1. The sort/limit/filter from the join_relationship weren't being added to the through_query at all
2. When the through_query is wrapped in a subquery, the parent correlation needs to be inside the subquery so the limit applies per-parent

This is related to https://github.com/ash-project/ash/pull/2486 (hence why some of the code is the same) but for Postgres, which uses lateral joins.

(Claude was used in investigating and fixing this bug)

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -1243,6 +1243,14 @@ defmodule AshPostgres.DataLayer do
         })
         |> Ash.Query.set_context(through_relationship.context)
         |> Ash.Query.do_filter(through_relationship.filter)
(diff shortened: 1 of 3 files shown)


zachdaniel:
`thanks

Issue #672, opened 2026-01-06
detect postgres v18+, to set different defaults and extensions


lucavenir:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Is your feature request related to a problem? Please describe.

Currently, `ash_postgres` generates a custom `uuid_generate_v7()` function, which is effectively useless when it comes to postgres v18+.

The main pain point is that trying to map a `uuid_v7_primary_key` in our domain to `uuidv7()` in postgres data layer can be cumbersome, as highlighted in #670.

### Describe the solution you'd like

I'd love `ash_postgres` not to generate the custom `uuid_generate_v7()` function, and to default mapping `uuid_v7_primary_key` to `uuidv7()` without extra effort.

### Describe alternatives you've considered

I'm currently writing:
```elixir
    migration_defaults id: "fragment(\"uuidv7()\")"
```

In each and every resource. Which is the cumbersome part. As explained [here](https://github.com/ash-project/ash_postgres/issues/670#issuecomment-3712612374), there's no "neat" way to globally override a migration default (which is good, if you ask me), so I'm sticking to this workaround until this is approved and implemented.

### Additional context

Thank you Zach for the ecosystem and the help you've given me on discord and in other issues ❤️

febarnett3:
Hi, I'm an ash-fellow. This looks like a good first issue for me to take on. I'll start working on it.

matt-savvy:
> Hi, I'm an ash-fellow. This looks like a good first issue for me to take on. I'll start working on it.

The main thing described here was updated by https://github.com/ash-project/ash_postgres/pull/674. Before you start working on any code changes, you might want to put together a list of what other defaults or extensions (if any) need to be set.

zachdaniel:
Yeah there may not be any more?

zachdaniel:
Sorry its not cleaned up in the issues list of so, but working to keep the project clean is still very valuable work!

lucavenir:
@febarnett3 howdies! it looks like this issue is kinda "closable", as AFAIK ash postgres supports v7 by default. the only bit left is the generation of that (now) useless function - maybe that's worth addressing?

zachdaniel:
@lucavenir I'll close this, but open to having that other thing fixed yes!

PR #674, opened 2026-01-17
Update MigrationGenerator, use uuidv7 when available

matt-savvy:
https://github.com/ash-project/ash_postgres/issues/672

Other changes needed
- include the `AshFunctions` `uuid_generate_v7` function on initial migration only on Postgres versions where it's not available
- figure out if there's any place where migrations are generated for going from one major version to the other. if downgrading to 16, we would need to add `uuid_generate_v7`, if upgrading to 18, remove it. 
- update docs

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/documentation/topics/development/upgrading-to-postgres-18.md
+++ b/documentation/topics/development/upgrading-to-postgres-18.md
@@ -1,3 +1,9 @@
+<!--
+SPDX-FileCopyrightText: 2020 Zach Daniel
+
+SPDX-License-Identifier: MIT
+-->
+
 # Using Postgres 18
 
 Postgres 18 adds built-in functions for working with version 7 (time-ordered) UUIDs.


matt-savvy:
> Yeah the big question will be how we make this smooth for those upgrading from 17-> 18 for example. 

Correct me if I'm wrong, but there's no specific Igniter task or anything that people would run when upgrading DB versions, right? I haven't had to do this on any Ash projects yet. I imagine the process is just 
1. update DB and data as normal
2. update the `min_pg_version` in the `Repo` 

> We may need to add an option for example for folks to retain the old behavior when the upgrade, and write a guide here.

What do you have in mind? Just give people a way so that they can update their `min_pg_version` while opting out of any migrations being generated for these fields? If so, would you want a new callback added added to the `AshPostgres.Repo` behavior? Or did you have something else in mind? (another option in the macro, flag / env var for the `ash.codegen` task? Something else? 

Also, are you thinking that this option would exist long term? Or just for a "deprecation" period of time for existing projects?

zachdaniel:
It would likely exist long term, and would be something like this:

```elixir
def use_builtin_uuidv7_function?(), do: true
```

and the default of that function would depend on what `min_pg_version` returned.

zachdaniel:
and correct, no task exists for that currently.

Issue #610, opened 2025-09-03
Creating indexes concurrently fails for schema-driven multi-tenancy


smt116:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

Elixir 1.18.3-otp-27
Erlang 27.3.3

ash: "3.5.37
ash_admin: "0.13.18
ash_ai: "0.2.12
ash_appsignal: "0.1.3"
ash_archival: "2.0.1"
ash_authentication_phoenix: "2.10.5"
ash_cloak: "0.1.7"
ash_json_api: "1.4.42
ash_oban: "0.4.12
ash_paper_trail: "0.5.6"
ash_phoenix: "2.3.16
ash_postgres: "2.6.17
ash_sql: "0.2.92
ash_state_machine: "0.2.12

### Operating system

macOS

### Current Behavior

Creating index concurrently with `@disable_ddl_transaction true` and `@disable_migration_lock true` fails with the following exception if using schema-driven tenancy:

```
* ** (Postgrex.Error) ERROR 25001 (active_sql_transaction) CREATE INDEX CONCURRENTLY cannot run inside a transaction block
  (ecto_sql 3.13.2) lib/ecto/adapters/sql.ex:1098: Ecto.Adapters.SQL.raise_sql_call_error/1
  (elixir 1.18.3) lib/enum.ex:1714: Enum."-map/2-lists^map/1-1-"/2
  (ecto_sql 3.13.2) lib/ecto/adapters/sql.ex:1219: Ecto.Adapters.SQL.execute_ddl/4
  (ecto_sql 3.13.2) lib/ecto/migration/runner.ex:348: Ecto.Migration.Runner.log_and_execute_ddl/3
  (elixir 1.18.3) lib/enum.ex:1714: Enum."-map/2-lists^map/1-1-"/2
  (elixir 1.18.3) lib/enum.ex:1714: Enum."-map/2-lists^map/1-1-"/2
  (ecto_sql 3.13.2) lib/ecto/migration/runner.ex:311: Ecto.Migration.Runner.perform_operation/3
  (stdlib 6.2.2) timer.erl:595: :timer.tc/2
  (ecto_sql 3.13.2) lib/ecto/migration/runner.ex:25: Ecto.Migration.Runner.run/8
[...]

PR #685, opened 2026-02-04
Fix concurrent index migrations with schema-driven multitenancy #610

RylandBangerter85:
- Detect migrations with @disable_ddl_transaction attribute
- Run concurrent index migrations outside transaction context
- Use Ecto.Adapters.SQL.checkout to get fresh connection
- Add comprehensive test suite for concurrent index scenarios
- Maintain backward compatibility with regular migrations

Fixes issue where CREATE INDEX CONCURRENTLY fails when creating tenants

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -2375,8 +2375,27 @@ defmodule AshPostgres.DataLayer do
 
     fields_to_upsert =
       case fields_to_upsert do
-        [] -> keys
-        fields_to_upsert -> fields_to_upsert
+        [] ->
+          keys
+
+        fields_to_upsert ->
+          # Include fields with update_defaults (e.g. update_timestamp)
+          # even if they aren't in the changeset attributes or upsert_fields.
+          # These fields should always be refreshed when an upsert modifies fields.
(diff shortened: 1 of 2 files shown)


lib/multitenancy.ex
```diff
@@ -123,4 +132,55 @@ defmodule AshPostgres.MultiTenancy do
   defp tenant_name_regex do
     ~r/^[a-zA-Z0-9_-]+$/
   end
+
+  # Check if a migration requires no transaction by examining the file content
+  # and the migration module attributes
+  defp migration_requires_no_transaction?(file, mod) do
+    # First check the file content for @disable_ddl_transaction
+    file_content = File.read!(file)
+    has_disable_ddl_transaction? = String.contains?(file_content, "@disable_ddl_transaction")
```
zachdaniel: Do we really need to check both? If the module is compiled we ought to be able to just ask the module and not bother looking at the file, since they could *theoretically* do `Module.put_attribute(__MODULE__, :disable_ddl_transaction, true)`

Issue #696, opened 2026-02-19
bulk_create with upsert? does not update update_timestamp fields on conflict


mbaertschi:
`Ash.bulk_create/4` with `upsert?: true` never updates `update_timestamp` attributes (e.g. `updated_at`) on conflict, even when the upsert modifies other fields. The `updated_at` column retains the value from the original insert.

## Expected behavior

`update_timestamp` fields should always be refreshed when a record is upserted, regardless of whether the field is listed in `upsert_fields`. This is consistent with the semantics of `update_timestamp` — the record was modified, so its update timestamp should reflect that.

By contrast, `Ash.Changeset.set_on_upsert/2` (used by the ETS/Mnesia data layers for single-record upserts) handles this correctly — it unconditionally merges `upsert_update_defaults()` into the SET list.

## Root cause

In `AshPostgres.DataLayer.upsert_set/4`, the fields included in `ON CONFLICT DO UPDATE SET` are filtered against `attributes_changing_anywhere`:

```elixir
attributes_changing_anywhere =
  changesets |> Enum.flat_map(&Map.keys(&1.attributes)) |> Enum.uniq()

upsert_fields =
  (options[:upsert_fields] || []) |> Enum.filter(&(&1 in attributes_changing_anywhere))
```

`update_timestamp` attributes have `writable?: false` by default, so they never appear in `changeset.attributes` and are filtered out. The code downstream already handles `update_defaults` correctly (via `COALESCE(EXCLUDED.field, default)`) — but the field never reaches that path.

## Suggested fix

In `upsert_set/4`, always include fields that have an `update_default` in `fields_to_upsert`, independent of `attributes_changing_anywhere` and `upsert_fields`.

## Reproduction

```elixir
defmodule MyApp.Thing do
  use Ash.Resource, data_layer: AshPostgres.DataLayer

  postgres do
    table "things"
    repo MyApp.Repo
  end

  actions do
    defaults [:read, :destroy]
    create :create do
      primary? true
      accept [:name]
    end
  end

  attributes do
    uuid_primary_key :id
    attribute :name, :string, allow_nil?: false, public?: true
    create_timestamp :created_at
    update_timestamp :updated_at
  end
end

# Initial create
Ash.bulk_create([%{name: "hello"}], Thing, :create, upsert?: false)

# Upsert with changed data
Ash.bulk_create(
  [%{name: "world"}],
  Thing,
  :create,
  upsert?: true,
  upsert_fields: [:name]
)

[...]

PR #697, opened 2026-02-20
fix: bulk_create with upsert now updates update_timestamp fields

mbaertschi:
## Summary

- `update_timestamp` attributes (e.g. `updated_at`) were never included in the
  `ON CONFLICT DO UPDATE SET` clause because they have `writable?: false` and
  thus never appear in `changeset.attributes`. This ensures fields with
  `update_defaults` are always included when an upsert modifies fields.
- Adds a `touch_update_defaults?` data layer context option to opt out of this
  behavior when needed:

  ```elixir
  Ash.bulk_create(inputs, Resource, :create,
    upsert?: true,
    upsert_fields: [:name],
    context: %{data_layer: %{touch_update_defaults?: false}}
  )
  

Closes ash-project/ash_postgres#696


# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -2382,12 +2382,20 @@ defmodule AshPostgres.DataLayer do
           # Include fields with update_defaults (e.g. update_timestamp)
           # even if they aren't in the changeset attributes or upsert_fields.
           # These fields should always be refreshed when an upsert modifies fields.
-          update_default_fields =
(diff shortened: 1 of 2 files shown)


zachdaniel:
It can set that context automatically, and would work for ETS for example.

PR #701, opened 2026-02-25
refactor: read touch_update_defaults? from options instead of changeset context

mbaertschi:
Read the touch_update_defaults? flag from the bulk_create options map
rather than from changeset context.

Depends on ash-project/ash#2590 which adds touch_update_defaults? to the
options passed to data layers. Only merge once that PR is merged and released.

Also support touch_update_defaults? option in non-bulk upsert path

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [x] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -3225,7 +3225,7 @@ defmodule AshPostgres.DataLayer do
   end
 
   @impl true
-  def upsert(resource, changeset, keys, identity \\ nil) do
+  def upsert(resource, changeset, keys, identity) do
     if AshPostgres.DataLayer.Info.manage_tenant_update?(resource) do
       {:error, "Cannot currently upsert a resource that owns a tenant"}
     else


mbaertschi:
@zachdaniel tests will pass once https://github.com/ash-project/ash/pull/2590 is merged into ash

lib/data_layer.ex
```diff
@@ -2382,9 +2382,8 @@ defmodule AshPostgres.DataLayer do
           # Include fields with update_defaults (e.g. update_timestamp)
           # even if they aren't in the changeset attributes or upsert_fields.
           # These fields should always be refreshed when an upsert modifies fields.
-          # Can be disabled via context: %{data_layer: %{touch_update_defaults?: false}}
-          touch_update_defaults? =
-            Enum.at(changesets, 0).context[:data_layer][:touch_update_defaults?] != false
+          # Can be disabled via touch_update_defaults?: false option
+          touch_update_defaults? = Map.get(options, :touch_update_defaults?, true)
```
zachdaniel: I think it would be good to check *both* here so that its backwards compatible.

mbaertschi: done

zachdaniel:
Same for the `ash_sqlite` version 🙇

PR #702, opened 2026-02-26
test: regression tests for bulk manage_relationship rollback bug

barnabasJ:
## Summary

- Adds regression tests for a bug where `run_after_action_hooks` in bulk actions does not call `Ash.DataLayer.rollback` when `manage_relationships` fails
- Tests cover `bulk_create` (including upserts), `bulk_update`, and `bulk_destroy`
- All bulk tests use `batch_size: 2` with 5 inputs to exercise batch boundary behavior
- Tests cover all strategies (`[:stream]`, `[:atomic, :stream]`, `[:atomic_batches, :stream]`) and both transaction modes (`:batch` and `:all`)

## Related

- Fix PR: ash-project/ash#2592

> **Note:** The 7 `transaction: :batch` tests will fail until the fix in ash-project/ash#2592 is merged and the ash dependency is updated. The remaining 12 tests (`:all` + controls) pass on the current codebase.

## Destroy path inconsistency

Worth noting: the non-bulk `Ash.destroy!/2` control test behaves differently from `Ash.create!/2` and `Ash.update!/2`. Single create and update both exercise `manage_relationship` and correctly roll back when the child validation fails. Single destroy does **not** exercise `manage_relationship` for `has_many` with `type: :create` at all — the parent is simply deleted without attempting child creation. The bulk destroy path **does** exercise it (via `run_after_action_hooks`), so the bug manifests there.

This means there's an inconsistency between the single and bulk destroy paths regarding when `manage_relationship` is evaluated. It may be intentional (managing relationships on destroy is unusual), but it's worth investigating whether the single destroy path should also honor `manage_relationship` changes, or whether the bulk destroy path should skip them to match.

## Test plan

- [ ] Without fix (stock ash): `mix test test/bulk_manage_relationship_rollback_test.exs` → 19 tests, 7 failures (all `:batch`)
- [ ] With fix (`ASH_VERSION=local`): 19 tests, 0 failures

### DIFF
--- a/test/bulk_manage_relationship_rollback_test.exs
+++ b/test/bulk_manage_relationship_rollback_test.exs
@@ -24,7 +24,7 @@ defmodule AshPostgres.BulkManageRelationshipRollbackTest do
   """
   use AshPostgres.RepoCase, async: false
 
-  alias AshPostgres.Test.{RollbackParent, RollbackChild}
+  alias AshPostgres.Test.{RollbackChild, RollbackParent}
 
   # ============================================================================
   # bulk_create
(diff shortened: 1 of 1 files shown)


zachdaniel:
> It may be intentional (managing relationships on destroy is unusual), but it's worth investigating whether the single destroy path should also honor manage_relationship changes, or whether the bulk destroy path should skip them to match.

It is strange to manage_relationship on destroy but regardless bulk + non-bulk semantics should be identical. Would you mind opening a separate issue about that? Great catch 🙇

zachdaniel:
> Sure, is the expectation that manage_relationship is just not supported by destroy actions?

It's meant to work roughly the same, especially considering that destroy actions could actually be `soft? true` mapping to an update action. But in general we don't require that the source record still exist, except certain configurations would naturally raise an error (like `relate`).

PR #748, opened 2026-05-13
Resource generator: many-to-many, views, no-PK tables, and prompt UX

weljoda:
# Contributor checklist

- [ ] Chores
- [x] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

This PR improves `mix ash_postgres.gen.resources` adding many_to_many detection, views and better conflict resolution. See: https://discordapp.com/channels/711271361523351632/1495458363960721629

I also set up an example project with a legacy DB schema, to test the generator on a real-world schema. https://github.com/weljoda/gen_resources_example

## New capabilities

- many_to_many detection. When a table is a pure join table (two FKs to two distinct tables, identity backed by either a composite primary key or a unique index over both FK columns), the generator now emits a many_to_many on each endpoint alongside the has_many to the join resource.
- PostgreSQL VIEW / MATERIALIZED VIEW support via `--include-views`. View-backed resources are emitted with `migrate? false` and a :read-only default action. The underlying introspection queries were ported from information_schema to pg_catalog so views are discoverable at all.
- Tables without a primary key are now generated with `require_primary_key? false` and a `:read, create: :*` default action, each carrying a # WARNING: comment pointing the user at the fix.
- `--skip-many-to-many` flag for users who prefer just has_many on join tables.

## Conflict-resolution UX overhaul

When multiple FKs produce duplicate default relationship names, the generator now:

- Shows full conflict context (resource, type, destination, FK column, constraint name) and a suggests an alternative name when common naming patterns are detected.
- Offers a menu with the following options: use suggested, auto-resolve remaining on this resource, skip this relationship.
- Tracks a [seen/total] progress counter so you know how many prompts are left per resource.
- Resolves names against existing attributes and previously named relationships, recursing until no collisions remain.

In my usecase this reduced the prompts by using auto-resolve to 4 from 58 previously.

## Bug fixes folded in

- opts[:yes] was being checked as opts[:yes?], so --yes never actually short-circuited the conflict prompts. Fixed.
- add_primary_key/2 was called with a boolean instead of the attribute struct, so its pattern match never fired and composite-PK join tables. Previously the compiler raised an error when generating resources with composite-PK join tables.

## Generated-code ordering

- relationships do now groups by type (belongs_to → has_one → has_many → many_to_many) then alphabetical within each group, instead of all-alphabetical with many_to_many interspersed.
- references do is alphabetical by relationship name, matching the belongs_to order rather than FK-insertion order.

### DIFF
--- a/lib/mix/tasks/ash_postgres.gen.resources.ex
+++ b/lib/mix/tasks/ash_postgres.gen.resources.ex
@@ -35,6 +35,7 @@ if Code.ensure_loaded?(Igniter) do
     - `no-migrations` - Do not generate snapshots & migrations for the resources. Defaults to `false`.
     - `skip-unknown` - Skip any attributes with types that we don't have a corresponding Elixir type for, and relationships that we can't assume the name of.
     - `fragments` - Generate attributes and relationships in a separate fragment file. This allows the fragment to be regenerated without affecting user customizations in the main resource file. Defaults to `false`.
+    - `skip-many-to-many` - Do not generate `many_to_many` relationships for detected join tables. `has_many` relationships to join tables are still generated. Defaults to `false`.
 
     ## Tables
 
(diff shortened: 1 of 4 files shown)


zachdaniel:

PR #752, opened 2026-05-14
fix: handle removal of :context multitenancy in migrations

Ryan6794:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

When a resource transitions from :context multitenancy to global?: true (or removing multitenancy entirely), the migration generator produces an invalid migration that tries to drop a foreign key constraint that doesn't exist yet.

# Bugs:

1. Wrong snapshot folder lookup - :context snapshots are stored in tenants/ but after removing multitenancy the code looks in the base folder. The old snapshot is not found so the table is treated as brand new.

2. Incorrect FK drop generated - has_reference?/2 did not account for the case where both the table and the referenced resource previously had :context multitenancy. Since the FK was never created in the global schema, it should not be dropped.

# Fix: 
1. get_existing_snapshot/2: fall back to checking the alternate folder when the primary lookup returns nil.
2. has_reference?/2: return false when both multitenancy strategies are :context, since no global FK exists to drop.

# Testing:
Reproduced using the demo repo linked in the issue: https://github.com/serpent213/ash_postgres_multitenancy_migration_demo

After the fix, all three steps complete successfully:
1. mix ash.reset
2.  mix ash_postgres.generate_migrations --dev
3.  mix ash.migrate

### DIFF
--- a/lib/migration_generator/migration_generator.ex
+++ b/lib/migration_generator/migration_generator.ex
@@ -3109,32 +3109,22 @@ defmodule AshPostgres.MigrationGenerator do
     not is_nil(Map.get(attribute, :references)) and
       !(attribute.references.multitenancy &&
           attribute.references.multitenancy.strategy == :context &&
-          (is_nil(multitenancy) || multitenancy.strategy == :attribute))
+          (is_nil(multitenancy) || multitenancy.strategy == :attribute)) and
+      !(multitenancy && multitenancy.strategy == :context &&
+          attribute.references.multitenancy &&
+          attribute.references.multitenancy.strategy == :context)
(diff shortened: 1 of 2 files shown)


zachdaniel:
Could we add a test for this too?

zachdaniel:
I think we're going to need to add a interactive prompt i.e

`are you renaming tenant.table to table` or `are you moving ...` because there isn't really a way to tell deterministically otherwise.

We'd also want a test case of what happens when a same table already exists in both tenant migrations and non-tenant migrations

PR #757, opened 2026-05-20
fix: Split unmanaged & managed resources and orphan detection

### LINKED ISSUE #756: migrate? false makes codegen want to delete the resource
### Code of Conduct

### AI Policy

### Versions

```
elixir --version
Erlang/OTP 28 [erts-16.2] [source] [64-bit] [smp:32:32] [ds:32:32:10] [async-threads:1] [jit:ns]

Elixir 1.19.4 (compiled with Erlang/OTP 28)
```

This is on ash_postgres 2.9.1

### Operating system

Linux

### Current Behavior

When an ash resource is created without a migration via the `migrate? false` option, and the `ash_postgres.generate_migrations` is ran, the result is that the task says the following:

```
Table project_statuses no longer has a resource. Generate a migration to DROP this table? This will permanently remove the table and its data. [Yn]
```

On ash_postgres 2.8, this does not happen.

### Reproduction

Reproduction doesn't have a clean answer. It appears to require a specific configuration that might be largely unique to us.

This is only impacting resources from a repo whose schema was being managed by a different application. 

The reproduction is really awkward to write as a test case but this is what it looks like:

* Have a pre-existing table in a schema you load up as a part of your release strategy. We have a whole little setup to accomplish this.
* Write Ash resources to correspond to that table
* Disable migrations with `migrate? false` as a part of the resource's `postgres` configuration block.
* Using ash_postgres 2.9.0+, run the codegen
* You will be told that the table has no resource

Selecting "no" on the prompt appears to be respected, so this has a workaround but the prompt made it sound like there was a far deeper issue with the resource configuration that would detonate in an unexpected way.

I fully admit that this is not a standard configuration. We're working to kill it.

### Expected Behavior

The expectation is that 2.9.x would behave like 2.8.x in that tables without migrations would not be told there are no corresponding resources when the resources do exist.

ChivukulaVirinchi:
migrate? false dropped unmanaged tables too eagerly. 

Fixed it to split unmanaged and managed resources,and pass unmanaged resources to orphan detection.

So final current_table_keys becomes the proper managed U unmanaged.

Fixes #756 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/migration_generator/migration_generator.ex
+++ b/lib/migration_generator/migration_generator.ex
@@ -32,12 +32,22 @@ defmodule AshPostgres.MigrationGenerator do
 
     all_resources = Enum.uniq(Enum.flat_map(domains, &Ash.Domain.Info.resources/1))
 
-    {tenant_snapshots, snapshots} =
+    postgres_resources =
       all_resources
       |> Enum.filter(fn resource ->
-        Ash.DataLayer.data_layer(resource) == AshPostgres.DataLayer &&
-          AshPostgres.DataLayer.Info.migrate?(resource)
+        Ash.DataLayer.data_layer(resource) == AshPostgres.DataLayer
       end)
+
+    {managed_resources, unmanaged_resources} =
+      Enum.split_with(postgres_resources, &AshPostgres.DataLayer.Info.migrate?/1)
+
+    {tenant_snapshots, snapshots} =
+      managed_resources
+      |> Enum.flat_map(&get_snapshots(&1, all_resources))
+      |> Enum.split_with(&(&1.multitenancy.strategy == :context))
+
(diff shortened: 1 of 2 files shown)


zachdaniel (approved):

zachdaniel:
Let's address the reuse compliance issues and any other failing checks. The audit one can be ignored.

ChivukulaVirinchi:
The failing checks are because of an already failing branch that I started work off of, which was in master. 
Which commit should I rebase them off of?

zachdaniel:
I'll sort it out in main 😄

Issue #731, opened 2026-04-11
Generating a migration after setting a schema to a Resource re-creates the Resource's table instead of just setting the schema


gcugnet:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

| Component | Version |
|-----------|---------|
| Elixir | 1.19.2 |
| Erlang/OTP | 28 |
| Ash | 3.23.1 |
| AshPostgres | 2.8.0 |

### Operating system

NixOS 25.11

### Current Behavior

In a Ash Resource, when adding the attribute `schema` in the `postgres` block, and then running `mix ash.codegen`, it generates a migration which creates a brand new table rather than just changing the schema in the table definition.

### Reproduction

1. create a new Ash Resource:
    ```elixir
    defmodule Prepair.Catalog.NewModel do
      use Ash.Resource,
        domain: Prepair.Catalog,
        data_layer: AshPostgres.DataLayer
    
      postgres do
        repo Prepair.Repo
        table "new_models"
      end
    
      attributes do
        uuid_primary_key :id, public?: true
        attribute :name, :string, public?: true
        create_timestamp :inserted_at, type: :utc_datetime
        update_timestamp :updated_at, type: :utc_datetime
      end
    end
    ```
2. run `mix ash.codegen`. It generates a file with a migration to create the table. Great!
3. modify the Ash Resource you previously created, just to add a `schema` in the `postgres` definition block:
    ```elixir
    [...]

      postgres do
        repo Prepair.Repo
        schema "catalog"
        table "new_models"
      end

    [...]
    ```
4. run `mix ash.codegen` again. It generates a new migration file, with an unexpected behaviour: table re-creation instead of modification:
    ```elixir
    defmodule Prepair.Repo.Migrations.MoveNewModelsToCatalogSchema do
      @moduledoc """
      Updates resources based on their most recent snapshots.
    
      This file was autogenerated with `mix ash_postgres.generate_migrations`
      """
    
      use Ecto.Migration
    
      def up do
        execute("CREATE SCHEMA IF NOT EXISTS catalog")
      
        create table(:new_models, primary_key: false, prefix: "catalog") do
          add :id, :uuid,
            null: false,
            default: fragment("gen_random_uuid()"),
            primary_key: true
        
          add :name, :text
        
          add :inserted_at, :utc_datetime,
            null: false,
[...]

PR #758, opened 2026-05-21
Fix/move table schema migrations

febarnett3:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

**Summary**

Resolves issue https://github.com/ash-project/ash_postgres/issues/731
Fixes schema-only resource changes being generated as `drop table` + `create table`.

The generator matches snapshots by `{table, schema}`, so changing only `schema "..."` made the old table look removed and the new schema table look newly added. This PR detects same-table/different-schema changes as schema moves and emits `ALTER TABLE ... SET SCHEMA ...` instead, preserving existing data.

**Changes**

- Added `MoveTableSchema` operation.
- Reconnects old/new snapshots when a table moves schemas.
- Prevents orphaned old-schema snapshots from becoming table drops.
- Ensures schema moves run before later changes against the new schema.
- Avoids unnecessary FK drop/recreate when a referenced table moved schemas.
- Tests

**Added test coverage for:**

- Adding a schema to an existing table.
- Moving from one schema to another.
- Moving schema and adding a column in the same migration.
- Moving a referenced table without recreating FKs.

### DIFF
--- a/lib/migration_generator/migration_generator.ex
+++ b/lib/migration_generator/migration_generator.ex
@@ -489,10 +489,25 @@ defmodule AshPostgres.MigrationGenerator do
         Enum.map(deduped, fn {%{table: table, schema: schema} = snapshot, existing_snapshot} ->
           cond do
             move_info = Map.get(schema_move_map, {table, schema}) ->
-              {snapshot, move_info.snapshot}
+              if move_info[:declined?] do
+                {snapshot, nil}
+              else
+                {snapshot, move_info.snapshot}
+              end
 
             rename_info = Map.get(rename_map, {table, schema}) ->
               {snapshot, rename_info.snapshot}
+
+            existing_snapshot && existing_snapshot.schema != schema ->
+              if moving_table_to_schema?(
+                   {table, existing_snapshot.schema},
+                   {table, schema},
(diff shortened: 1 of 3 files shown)


zachdaniel:
I think we may need to ask if you are renaming the schema or not interactively, similar to how we do with other renames. The reason for this is that I think there could be some weirdness here if you have the same table but in different schemas. We may want some tests around those scenarios as well.

febarnett3:
That's a good point. I agree this should probably mirror the table rename behavior. The current logic assumes one same-table/different-schema candidate means a schema move, but that could be ambiguous if an unrelated table with the same name is added in another schema. I can add an interactive confirmation and tests around that case.

febarnett3:
Added the suggested prompt handling for inferred schema moves. If the user declines the schema move, the generator treats the target-schema table as new and falls back to the existing orphan-table drop prompt for the old table, allowing the same table name to intentionally exist in different schemas.
I also guarded the deduped snapshot matching path so a declined schema move cannot still generate `ALTER TABLE ... SET SCHEMA`, and added regression tests for the decline behavior.

PR #765, opened 2026-05-27
fix: Filter out empty snapshot directories

Rutgerdj:
This fixes an issue where you would not get the prompt to rename a table when there's already a directory with the same name as the new table name.

I ran into this issue after reverting some resource snapshots using git, which removed the files, but left the (empty) folders in place. 
Probably not a super common scenario, but good to have covered imo. 

Added a regression test that verifies the behaviour.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


Rutgerdj:
Failing tests are also failing on main it seems. 
At least the migration_generator tests are all green

zachdaniel:

PR #768, opened 2026-05-28
fix: add dump_to_native fallback for migration defaults when safe

### LINKED ISSUE #527: `ash_postgres.generate_migrations` not `dumping_to_native` when generating a default value
**Describe the bug**
When using a custom type for an attribute with a default value the migration is generated with the `internal` value rather than the `dump_to_native` value

**To Reproduce**
A minimal set of resource definitions and calls that can reproduce the bug.

**Expected behavior**
The generated migration to have the default value as the one that gets returned from `dump_to_native`

** Runtime
- elixir 1.18.1-otp-27
- erlang 27.1.3
 - macOS 15.5
 - Ash 3.4.68
 - Ash Postgres 2.5.11

**Additional context**

CaydenLords:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Fixes #527


lib/migration_generator/migration_generator.ex
```diff
@@ -4479,7 +4479,13 @@ defmodule AshPostgres.MigrationGenerator do
             default
 
           :error ->
-            EctoMigrationDefault.to_default(value)
+            case migration_default_from_dump(type, Map.get(attr, :constraints, []), value) do
```
zachdaniel: 🤔 looking at this, I think this may unfortunately be a breaking change. I actually don't think we can/should do this. We have a callback available in `AshPostgres.Type` now like so:

```elixir
  @callback value_to_postgres_default(Ash.Type.t(), Ash.Type.constraints(), term) ::
              {:ok, String.t()} | :error

``` 

This may not have been available when the person submitted that issue. So I think instead we should update our documentation in an appropriate place to explain this callback and then close the corresponding issue. Would you mind finding a good place to emphasize this callback?

zachdaniel (approved):

CaydenLords:
Alright, I'm moving this code to the fallback branch in my fork and opening a new one for the documentation.

PR #778, opened 2026-06-10
Add partial identity join sorting test

jechol:
## Summary

Adds a regression test for sorting a `many_to_many` relationship by a join resource attribute when the join resource uses a partial identity.

This covers the behavior fixed in ash-project/ash#2746.

The test uses a polymorphic join resource shape where multiple nullable foreign keys can exist on the same resource, but only the `post_id` / `user_id` pair is unique when both keys are present:

```elixir
identity :post_user, [:post_id, :user_id],
  where: expr(not is_nil(post_id) and not is_nil(user_id))
```

The loaded relationship then sorts through the join relationship field:

```elixir
sort: [Ash.Sort.expr_sort(parent(user_notification_recipients.order), :integer)]
```

## Notes

This test fixture mirrors polymorphic ordered join resources where a partial identity guarantees one join row for a specific source/destination pair.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/priv/test_repo/migrations/20260610180000_add_notification_recipients.exs
+++ b/priv/test_repo/migrations/20260610180000_add_notification_recipients.exs
@@ -1,0 +1,75 @@
+# SPDX-FileCopyrightText: 2019 ash_postgres contributors <https://github.com/ash-project/ash_postgres/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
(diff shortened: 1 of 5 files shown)


zachdaniel:

Issue #448, opened 2024-12-19
Specified timeout for an Ash action does not override the global timeout


Fasp96:
### **Describe the bug**
Encountered an issue where specifying a timeout for an Ash action does not override the default 15000ms timeout. Despite following the [Ash documentation on timeouts](https://hexdocs.pm/ash/timeouts.html#ways-to-specify-timeout), the action fails with the following error:

```elixir
[error] Postgrex.Protocol (#PID<0.442.0>) disconnected: ** (DBConnection.ConnectionError) client #PID<0.866.0> timed out because it queued and checked out the connection for longer than 15000ms
```

### **To Reproduce**
Example code:
```elixir
defp fetch_report_data(resource, report_filters, actor) do
  case resource
       |> Ash.read(actor: actor, timeout: :timer.minutes(5)) do
    {:ok, records} -> {:ok, resource, records}
    error -> error
  end
end

```

### **Steps to reproduce:**

Define an Ash resource.

Call the Ash.read/2 function with a custom timeout (:timer.minutes(5)).

Observe the timeout error despite specifying a longer timeout.

### **Expected behavior**
I expected the timeout: :timer.minutes(5) option in the Ash.read/2 function to increase the timeout duration for this specific call to 5 minutes.

### **Additional context**
I noticed that increasing the timeout globally in the `Ecto.Adapters.Postgres` configuration resolves the issue, but this is not desirable as I only want to increase the timeout for this specific action. The documentation mentions that the timeout option should work for individual calls, but it does not seem to be applied here.

zachdaniel:
Please see if adding `transaction? true` (I accidentally told you to try `transaction? false` in discord) to see if that is a workaround :)

torazar:
I'm going to work on this issue. I'd think the best fix is take the timeouts and have them be input into run_query. Would that work, or would you prefer the transaction? true fix?

zachdaniel:
That was mostly to gather information. I haven't thought about this in a while, so its hard to say if we need to pass it into `run_query` but generally speaking this should be handled *above* the data layer unless we are running transactions.

PR #779, opened 2026-06-16
Timeout override now works #448

torazar:
, and it won't simply default to the regular timeout if you changed that.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -420,6 +420,7 @@ defmodule AshPostgres.DataLayer do
 
   use Spark.Dsl.Extension,
     sections: @sections,
+    transformers: [AshPostgres.Transformers.AddQueryTimeoutPreparation],
     verifiers: [
       AshPostgres.Verifiers.PreventMultidimensionalArrayAggregates,
       AshPostgres.Verifiers.ValidateReferences,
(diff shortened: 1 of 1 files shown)


lib/data_layer.ex
```diff
@@ -420,6 +420,7 @@ defmodule AshPostgres.DataLayer do
 
   use Spark.Dsl.Extension,
     sections: @sections,
+    transformers: [AshPostgres.Transformers.AddQueryTimeoutPreparation],
```
zachdaniel, line 423: What's this? I don't think we have this preparation anywhere do we?

lib/data_layer.ex
```diff
@@ -833,7 +834,7 @@ defmodule AshPostgres.DataLayer do
 
   @impl true
   def set_context(resource, data_layer_query, context) do
-    AshSql.Query.set_context(resource, data_layer_query, AshPostgres.SqlImplementation, context)
+    AshPostgres.Sql.Query.set_context(resource, data_layer_query, context)
```
zachdaniel, line 837: I don't see this module anywhere either 🤔

torazar: That points to a module in /lib/sql.

lib/data_layer.ex
torazar: So, that creates a preparation step behind the scene to handle that.

lib/data_layer.ex
zachdaniel: There is no `lib/sql`?

torazar: I see... Guess I forgot to transfer it over. I'll do this merge over, I guess.

PR #782, opened 2026-06-25
Actually have the fix for #448.

torazar:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/mix/tasks/ash_postgres.generate_migrations.ex
+++ b/lib/mix/tasks/ash_postgres.generate_migrations.ex
@@ -8,7 +8,8 @@ defmodule Mix.Tasks.AshPostgres.GenerateMigrations do
 
   Options:
 
-  * `domains` - a comma separated list of Domain modules, for which migrations will be generated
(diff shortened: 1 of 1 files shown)


torazar:
Apparently I didn't know some things about git. Didn't realize I had to manually add new files I made.

zachdaniel:
Looking at the extent of this fix, it seems overbearing. Modifying all read actions on a resource, and threading through a timeout like this when in reality it seems like this issue should have a simple fix somewhere. Where is the timeout getting dropped from the query?

I think in this case the first thing we need is a diagnosis of exactly where the timeout issue is going wrong.

torazar:
Right. Turns out there are two places with problems. First, Ash never passes query.timeout to the data layer, Second, Ash only passes the Ecto query to run_query, not the Ash query.

zachdaniel:
So it would seem there is a minimal fix possible in that case?

PR #789, opened 2026-07-07
fix: resolve actor/context templates in through relationship filters

### LINKED ISSUE #788: `^context/1` and `^actor/1` templates are not supported in relationship filters used by `has_many ... through`
### Code of Conduct

### AI Policy

### Versions

ash_posgtres latest-main
erlang 27.1.2
elixir 1.20

### Operating system

MacOS Version 26.5.1

### Current Behavior

When the following `has_many` relationship includes a filter that uses `^context/1` (the same issue occurs with `^actor/1`):

```elixir
has_many :classrooms, AshPostgres.Test.Through.Classroom do
  public?(true)

  filter(
    expr(
      is_nil(^context(:sample_context)) or
        name == ^context(:sample_context)
    )
  )
end
```

and that relationship is used by a `has_many ... through` relationship:

```elixir
has_many :teachers, AshPostgres.Test.Through.Teacher do
  public?(true)
  through([:classrooms, :classroom_teachers, :teacher])
end
```

loading the `teachers` relationship fails.

Instead of resolving the `^context/1` (or `^actor/1`) template, AshPostgres attempts to compile the raw `{:_context, :sample_context}` expression into SQL and raises:

```text
** (RuntimeError) Unsupported expression in Elixir.AshPostgres.SqlImplementation query:
{:_context, :sample_context}
```

The error occurs while AshPostgres is building the SQL query for the `through` relationship.

### Reproduction

1. Clone the reproduction repository:

   `https://github.com/tunchamroeun/ash_postgres`

2. Run the failing test:

```sh
mix test test/through_relationships_test.exs:42
```

The test reproduces the error caused by using `^context/1` (or `^actor/1`) in a relationship filter that is referenced by a `has_many ... through` relationship.

### Expected Behavior

`^context/1` and `^actor/1` templates should be supported in relationship filter expressions, including when the relationship is used by a `has_many ... through` relationship.

The templates should be resolved before AshPostgres generates the SQL query, allowing the `through` relationship to load successfully.

barnabasJ:
## Problem

Fixes #788.

When a `has_many` relationship carries a filter that uses a `^context/1` or `^actor/1` template, and that relationship is used inside a `has_many ... through` relationship, loading the through relationship fails:

```
** (RuntimeError) Unsupported expression in Elixir.AshPostgres.SqlImplementation query:
{:_context, :sample_context}
```

The intermediate relationship's filter was applied via `Ash.Query.do_filter` **without filling the template**, so the raw template (`{:_context, :sample_context}`) was handed to the SQL implementation, which cannot compile it. The same class of bug affected `^actor/1`, `^arg/1` and `^tenant/0` templates.

## Fix

Fill `^actor`/`^context`/`^arg`/`^tenant` templates before the intermediate relationship filter is compiled into SQL, in both through code paths in `lib/data_layer.ex` (the recursive `build_through_queries/4` used for 3+ hop paths, and the inline path used for 2-hop / `many_to_many`).

This mirrors how a directly loaded relationship filter is hydrated in `Ash.Actions.Read.Relationships`: actor/args/tenant come from the source query, while the context is taken from the query the filter is being applied to (after its `set_context`). As a result, loading a `through` relationship resolves templates identically to loading each hop directly.

## Tests

Added intermediate relationships with `^context` and `^actor` filters on the `Through.School` test resource, plus `through` relationships using them, covering both the 3-hop (`build_through_queries`) and 2-hop (inline) paths. New tests in `test/through_relationships_test.exs` assert:

- the `^context`/`^actor` templates no longer raise, and
- the `^actor` template resolves to the actor's value and filters the through relationship correctly.

Full suite: 832 passing, `mix format` clean, `mix ash_postgres.generate_migrations --check` clean.

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -1306,13 +1306,28 @@ defmodule AshPostgres.DataLayer do
               start_bindings_at: through_binding
             })
             |> Ash.Query.set_context(Map.get(through_relationship, :context))
-            |> Ash.Query.do_filter(Map.get(through_relationship, :filter))
+            |> then(fn q ->
+              Ash.Query.do_filter(
+                q,
+                fill_relationship_filter_templates(
+                  Map.get(through_relationship, :filter),
+                  source_query,
+                  q
+                )
+              )
+            end)
             |> then(fn q ->
               # For through-list paths, the first relationship's filter applies to the through table
               # For many_to_many, the relationship filter is for the destination, not the through table
               if !is_atom(Map.get(relationship, :through)) ||
                    is_nil(Map.get(relationship, :through)) do
-                Ash.Query.do_filter(q, Map.get(relationship, :filter),
+                Ash.Query.do_filter(
+                  q,
+                  fill_relationship_filter_templates(
+                    Map.get(relationship, :filter),
+                    source_query,
+                    q
+                  ),
                   parent_stack: [relationship.source]
                 )
               else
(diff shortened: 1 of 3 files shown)


zachdaniel:

PR #790, opened 2026-07-09
fix: set tenant on destroy so exists() subqueries resolve against the tenant schema

### LINKED ISSUE #787: `destroy` doesn't set the tenant on the Ecto query — `exists()` in atomic validations resolves subqueries against the wrong schema under context multitenancy
### Code of Conduct

### AI Policy

### Versions

- ash_postgres 2.10.0 (also verified on 2.6.x and, on 2026-03-18, on then-current main)
- ash 3.29.3
- elixir 1.18.4 / OTP 27
- PostgreSQL 16

### Operating system

macOS 26 (Darwin 25.5.0

### Current Behavior

With context (schema) multitenancy, an atomic validation using `exists()` on a **destroy** action builds its subquery without the tenant prefix — the subquery targets the `public` schema instead of the tenant schema, failing with a relation-does-not-exist error (or silently reading the wrong schema if a same-named public table exists).

The identical `exists()` validation on an **update** action generates the correctly-prefixed subquery.

### Reproduction

Setup: any two resources with context (schema) multitenancy, e.g. `Parent` with `has_many :children`.

1. Give `Parent` a destroy action with an atomic validation using `exists()`:

   ```elixir
   destroy :destroy do
     validate absent(exists(children, true)), message: "has children"
   end

2. Ash.destroy!(parent, tenant: "tenant_a")

The exists subquery is generated without the "tenant_a" schema prefix (targets public."children"), failing with relation-does-not-exist — or matching the wrong tenant's rows if a same-named public table exists. The identical validation on an update action prefixes correctly.

Root cause (from reading lib/data_layer.ex in 2.10.0): update/2 pipes its query through set_tenant/3 and puts :tenant into __ash_bindings__ (added in 60ab568, v2.6.14); destroy/2 builds the analogous query with neither step, so exists() compiles with no tenant for the subquery prefix.

(I can turn this into a failing test case in the ash_postgres suite if useful.)

### Expected Behavior


`destroy/2` propagates the tenant to the Ecto query the same way `update/2` does, so `exists()` subqueries in atomic validations resolve against the tenant schema.

Happy to submit a PR mirroring the `update/2` pattern into `destroy/2` if that's the right layer for the fix — or point me at the preferred spot (`Ash.Actions.Helpers.add_context/2` not carrying `tenant` in the private context looked like a plausible alternative root).

The parenthetical in Reproduction covers their "ideally a failing test" ask honestly — we're offering it rather than pretending we have a repro repo.

Francisco-Castro:
## Problem

Fixes #787.

Under context (schema) multitenancy, an `exists()` subquery in a **destroy** action's atomic filter/validation compiles against the `public` schema instead of the tenant schema, raising `42P01 undefined_table` (or silently matching the wrong schema if a same-named public table exists). The identical `exists()` on an **update** action resolves correctly.

## Fix

`AshPostgres.DataLayer.destroy/2` hand-builds its Ecto query from `changeset.context`. `Ash.Actions.Helpers.add_context/2` seeds the private context with `actor`/`authorize?`/`tracer` but not the tenant, so when the destroy filter is compiled via `filter/3` → `AshSql.Join.related_subquery` (which reads `query.__ash_bindings__.context[:private][:tenant]`), the tenant is `nil` and the subquery falls back to `public`.

Update escapes this only because its `exists()` compiles through the atomics path (`__ash_bindings__.tenant`). Bulk `destroy_query/4` is unaffected — it routes through `Ash.Query.data_layer_query/2`, which already seeds `context.private.tenant`.

The fix propagates the tenant to the query (prefix + `__tenant__` via `set_tenant/3`) and seeds `context.private.tenant` into the bindings the same way `Ash.Query.data_layer_query/2` does for reads/bulk.

## Tests

Added `destroy(:destroy_with_exists)` on the multitenancy `Post` test resource (mirroring the existing `update_with_policy`) and a regression test in `test/multitenancy_test.exs`. Confirmed red (`42P01` against `public.multitenant_posts`) before the fix, green after.

Full suite: 850 tests + 2 doctests + 2 properties, 0 failures. `mix format` and `mix ash_postgres.generate_migrations --check` clean.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -4142,6 +4142,28 @@ defmodule AshPostgres.DataLayer do
         changeset.context
       )
       |> pkey_filter(record)
+      |> then(fn query ->
+        if changeset.tenant do
+          query =
+            set_tenant(resource, query, changeset.tenant)
+            |> elem(1)
+
+          Map.update!(query, :__ash_bindings__, fn bindings ->
+            Map.update(
+              bindings,
+              :context,
+              %{private: %{tenant: changeset.tenant}},
+              fn context ->
+                context
+                |> Map.put_new(:private, %{})
+                |> put_in([:private, :tenant], changeset.tenant)
+              end
+            )
+          end)
+        else
+          query
+        end
+      end)
 
     repo = AshSql.dynamic_repo(resource, AshPostgres.SqlImplementation, changeset)
 
--- a/test/multitenancy_test.exs
+++ b/test/multitenancy_test.exs
@@ -98,6 +98,21 @@ defmodule AshPostgres.Test.MultitenancyTest do
     |> Ash.update!()
   end
 
+  test "context multitenancy sets the tenant on exists() subqueries when destroying", %{
+    org1: org1
+  } do
+    post =
+      Post
+      |> Ash.Changeset.for_create(:create, %{name: "foo"}, tenant: org1)
+      |> Ash.create!()
+
+    post
+    |> Ash.Changeset.for_destroy(:destroy_with_exists, %{}, tenant: org1)
+    |> Ash.destroy!()
+
+    assert Post |> Ash.Query.set_tenant(org1) |> Ash.read!() == []
+  end
+
   test "attribute multitenancy is set on creation" do
     uuid = Ash.UUID.generate()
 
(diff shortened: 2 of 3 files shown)


zachdaniel:

PR #792, opened 2026-07-09
fix: warn when resources are skipped due to migrate?: false during mi…

### LINKED ISSUE #585: unable to generate migration manually after generating domain from tables
### Versions

- ash 3.5.25
- ash_admin 0.13.11
- ash_phoenix 2.3.9
- ash_postgres 2.6.9
- ash_sql 0.2.84
- igniter 0.6.10

```sh
$ elixir --version
Erlang/OTP 27 [erts-15.2.7] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit:ns]

Elixir 1.18.4 (compiled with Erlang/OTP 27)
```

### Operating system

nixos

### Current Behavior

```sh
> mix ash_postgres.gen.resources XIt.Authorization --no-migrations --tables roles,role_policies,policies,policy_groups
...
> mix ash_postgres.generate_migrations --domains XIt.Authorization --name xit_authorization
No changes detected, so no migrations or snapshots have been created.
```

sidenote: I tried this because running `mix ash_postgres.gen.resources` multiple times generates non unique migration names. causing duplicate migration exceptions when run.

### Reproduction

see above

### Expected Behavior

expected changes to be detected and a unique migration to be generated.

Jsime34:
## Problem

Fixes #585.

The issue here was that after generating the domain from the tables using mix ash_postgres.gen.resources, no unique migrations were being generated. This lead to cryptic errors (no migrations could be generated) to the user, who expected unique migration names to be generated. 

## Fix
After a lot of testing and investigation, I found out that there was no bug and the process was working as expected, as these resources had the atribbute migrate?(false). In order to let the user understand the situation they are in, I have chosen to enhance the "No changes detected" message, and let the user know about the migrate?(false) attribute. In addition to this, a regression test was added covering this new previously untested path.

## Tests

Passed the migrate tests (test.drop, test.create, test.migrate) and mix test. 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/migration_generator/migration_generator.ex
+++ b/lib/migration_generator/migration_generator.ex
@@ -95,7 +95,9 @@ defmodule AshPostgres.MigrationGenerator do
 
             Note: Some resources have `migrate?` set to `false` and were skipped.
 
-            If you expected migrations to be generated for them, remove `migrate?(false)` from their `postgres` block. Resources generated with `mix ash_postgres.gen.resources --no-migrations` have `migrate?` set to `false` by default.
+            If you expected migrations to be generated for them, remove `migrate?(false)`
+            from their `postgres` block. Resources generated with `mix ash_postgres.gen.resources
+            --no-migrations` have `migrate?` set to `false` by default.
             """)
           end
         end
--- a/lib/migration_generator/migration_generator.ex
+++ b/lib/migration_generator/migration_generator.ex
@@ -90,14 +90,10 @@ defmodule AshPostgres.MigrationGenerator do
               "No changes detected, so no migrations or snapshots have been created."
             )
           else
-            unmanaged_names =
-              unmanaged_resources
-              |> Enum.map_join(", ", &inspect/1)
-
             Mix.shell().info("""
             No changes detected, so no migrations or snapshots have been created.
 
-            Note: the following resources have `migrate?` set to `false` and were skipped: #{unmanaged_names}.
+            Note: Some resources have `migrate?` set to `false` and were skipped.
 
             If you expected migrations to be generated for them, remove `migrate?(false)` from their `postgres` block. Resources generated with `mix ash_postgres.gen.resources --no-migrations` have `migrate?` set to `false` by default.
             """)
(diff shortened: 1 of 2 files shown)


lib/migration_generator/migration_generator.ex
```diff
@@ -85,9 +85,23 @@ defmodule AshPostgres.MigrationGenerator do
...
+            )
+          else
+            unmanaged_names =
+              unmanaged_resources
+              |> Enum.map_join(", ", &inspect/1)
+
+            Mix.shell().info("""
+            No changes detected, so no migrations or snapshots have been created.
+
+            Note: the following resources have `migrate?` set to `false` and were skipped: #{unmanaged_names}.
```
zachdaniel: In many apps they have *a lot* of unmanaged resources. If we want to display them then at a minimum we should display them like:

```elixir
- Resource
- Resource
- Resource
```

or

```elixir
Resource,Resource2,Resource3
Resource4,Resource5,Resource6
```

etc.

However, I think that listing might be too much and it might be best to say something like: "Some resources were skipped due to having `migrate?` set to `false` ..."

Jsime34:
Alright I just fixed your suggestion, if there is any other observation on my code, please let me know.

lib/migration_generator/migration_generator.ex
```diff
@@ -85,9 +85,19 @@ defmodule AshPostgres.MigrationGenerator do
...
+            Mix.shell().info(
+              "No changes detected, so no migrations or snapshots have been created."
+            )
+          else
+            Mix.shell().info("""
+            No changes detected, so no migrations or snapshots have been created.
+
+            Note: Some resources have `migrate?` set to `false` and were skipped.
+
+            If you expected migrations to be generated for them, remove `migrate?(false)` from their `postgres` block. Resources generated with `mix ash_postgres.gen.resources --no-migrations` have `migrate?` set to `false` by default.
```
zachdaniel: Last step, should have called this out, but let's break up these sentences so they flow more naturally in the terminal. Keep them all about the same width as the above two sentences. After that its good to go!

Jsime34:
I've made the changes you requested.

zachdaniel:

PR #794, opened 2026-07-13
fix: return skipped upserts with nil identity values

jechol:
Fix single-record skipped-upsert reloads when an identity key is nil.

- encode nil identity values as `is_nil` filters
- add a regression test for `NULLS NOT DISTINCT` identities


zachdaniel:

PR #796, opened 2026-07-17
feat: allow custom indexes to exclude base filters

jechol:
## Summary

Adds an `include_base_filter?` option to PostgreSQL custom indexes.

It defaults to `true`, preserving the existing behavior. Setting it to `false` creates an index without merging the resource's `base_filter_sql` into its predicate.

## Motivation

Most custom indexes should inherit the resource base filter. Some indexes need to cover every row.

For example, when a referenced row is updated or deleted, PostgreSQL looks for matching rows in the referencing table. A partial index that excludes archived rows cannot support that lookup across all referencing rows.

Given:

```elixir
base_filter_sql "archived_at IS NULL"
```

a custom index can opt out of that predicate:

```elixir
custom_indexes do
  index [:account_id],
    name: "events_account_id_fkey_index",
    include_base_filter?: false
end
```

This generates:

```elixir
create index(
  :events,
  [:account_id],
  name: "events_account_id_fkey_index"
)
```

instead of:

```elixir
create index(
  :events,
  [:account_id],
  name: "events_account_id_fkey_index",
  where: "archived_at IS NULL"
)
```

## Compatibility

- Existing custom indexes continue to include the base filter.
- Older snapshots default to `include_base_filter?: true` when loaded.
- Newly generated snapshots include the option explicitly.
- Changing the option generates the required index rewrite.

## Tests

Added unit coverage confirming that `include_base_filter?: false` omits the resource base filter from the generated index.

# Contributor checklist

- [ ] Chores
- [X] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -23,6 +23,7 @@ spark_locals_without_parens = [
   identity_wheres_to_sql: 1,
   ignore?: 1,
   include: 1,
+  include_base_filter?: 1,
   index: 1,
   index: 2,
   index?: 1,
--- a/lib/custom_index.ex
+++ b/lib/custom_index.ex
@@ -21,7 +21,8 @@ defmodule AshPostgres.CustomIndex do
     :include,
     :nulls_distinct,
     :message,
-    :all_tenants?
+    :all_tenants?,
+    :include_base_filter?
   ]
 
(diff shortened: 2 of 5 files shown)


zachdaniel:

PR #798, opened 2026-07-18
improvement: replace pairwise operation ordering with a dependency-graph topological sort

jechol:
## Motivation

Migration operations were ordered by a ~69-clause pairwise `after?/2` comparator feeding an insertion sort. It's non-transitive and clause-order-dependent — two clause pairs turned out to be dead code, where a broader/earlier clause always shadowed a more specific one for the same struct pair.

It also had no way for an operation to declare "this depends on some fact being true yet." Global "always last" / "always first" tiers for things like `AddCheckConstraint`, `AlterDeferrability`, `RemovePrimaryKeyDown`, and `AddPrimaryKey` papered over what are really specific DDL ordering requirements with ad-hoc rules.

## What changed

Each operation now declares facts it `provides` and facts it `requires` (`operation_deps.ex`), and `toposort_operations/1` runs Kahn's algorithm over the resulting dependency graph.

A genuine cycle raises `OperationCycleError` instead of silently falling back to an arbitrary order — a later PR builds on this to let `custom_statements` opt into cross-table ordering, and a cycle there should fail loudly rather than produce a migration that runs in the wrong order.

`group_into_phases/3` and `clean_phases/1` are unchanged — they only consume the final order. `streamline/2` is nearly unchanged: its custom-statement guard (from 12cf97d8) is removed, since the structs it protected now carry `:schema` and the toposort can't place a same-table custom statement inside an add/alter window. `RenameTable`'s struct now uses the same generic `:table` field every other operation struct uses (previously `:old_table`/`:new_table`), and `RemovePrimaryKey` now carries its `:keys` so its FK-ordering requirement can be column-scoped.

## Bugs found during verification

Verified against a large real-world Ash application (~150 resources): a from-scratch schema regen applied to a real Postgres database, plus applying the same diff against a database restored from a production-sized backup, confirming the resulting schema matches in both cases. That surfaced two real bugs, both fixed here:

- A resource's own `schema` option defaults to `nil`, but `attribute.references.schema` is loaded as an explicit `"public"` string — building fact keys from these without normalizing silently dropped real cross-table FK ordering constraints.
- A unique index with a `where`/base filter can reference columns that aren't part of its keys (e.g. a soft-delete-scoped identity whose `where` checks `archived_at`, a column outside the index's own keys), with no structured way to know which ones without parsing raw SQL. `AddUniqueIndex` now conservatively requires every attribute added to the table first, but only when such a filter is actually present: applying that margin unconditionally would also force a self-referencing attribute (one whose own FK targets this same index, e.g. a self-referential `belongs_to`) to finish before the index it depends on — a genuine two-way cycle between that attribute and the index.

Tracing each fact's provider/consumer pairs also surfaced two pre-existing ordering bugs, fixed here with regression tests:

- Removing a check constraint and its column together could generate a `down` that recreates the constraint before re-adding the column it references. The dependency direction is now reversed: column removal/rename waits for the constraint's removal.
- Dropping a unique index still referenced by another table's foreign key was not ordered after the FK drop — Postgres blocks that drop for any unique constraint, not just the primary key.

## Testing

`test/migration_generator_test.exs`: all passing (two tests fixed for pre-existing whitespace fragility the new ordering exposed), plus new integration tests pinning the generated `up`/`down` order for the bugs above. New `test/migration_generator/operation_deps_test.exs` unit-tests the provides/requires model directly. Full `mix test`: 858 passed.

Writing up what each fact means (in response to review) surfaced facts that were carried over from the old `after?/2` rules but had no remaining consumer once traced through — each was removed individually and verified against the full suite. Every remaining fact has a documented provider/consumer pair and test coverage.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [X] Refactoring
- [ ] Update dependencies

### DIFF
--- a/lib/migration_generator/operation.ex
+++ b/lib/migration_generator/operation.ex
@@ -157,13 +157,13 @@ defmodule AshPostgres.MigrationGenerator.Operation do
 
   defmodule RenameTable do
     @moduledoc false
-    defstruct [:old_table, :new_table, :schema, :multitenancy, :repo, no_phase: true]
+    defstruct [:old_table, :table, :schema, :multitenancy, :repo, no_phase: true]
 
     import Helper, only: [as_atom: 1]
 
     def up(%{
           old_table: old_table,
-          new_table: new_table,
+          table: new_table,
           schema: schema,
           multitenancy: multitenancy
         }) do
@@ -175,7 +175,7 @@ defmodule AshPostgres.MigrationGenerator.Operation do
 
     def down(%{
           old_table: old_table,
-          new_table: new_table,
+          table: new_table,
           schema: schema,
           multitenancy: multitenancy
         }) do
@@ -1046,7 +1046,7 @@ defmodule AshPostgres.MigrationGenerator.Operation do
 
   defmodule AddCustomStatement do
     @moduledoc false
-    defstruct [:statement, :table, no_phase: true]
+    defstruct [:statement, :table, :schema, no_phase: true]
 
     def up(%{statement: %{up: up, code?: false}}) do
       """
@@ -1075,7 +1075,7 @@ defmodule AshPostgres.MigrationGenerator.Operation do
 
   defmodule RemoveCustomStatement do
     @moduledoc false
-    defstruct [:statement, :table, no_phase: true]
+    defstruct [:statement, :table, :schema, no_phase: true]
 
     def up(%{statement: statement, table: table}) do
       AddCustomStatement.down(%AddCustomStatement{statement: statement, table: table})
@@ -1307,7 +1307,7 @@ defmodule AshPostgres.MigrationGenerator.Operation do
 
   defmodule RemovePrimaryKey do
     @moduledoc false
-    defstruct [:schema, :table, no_phase: true]
+    defstruct [:schema, :table, :keys, no_phase: true]
 
     def up(%{schema: schema, table: table, old_multitenancy: multitenancy}) do
       cond do
--- a/lib/migration_generator/operation_cycle_error.ex
+++ b/lib/migration_generator/operation_cycle_error.ex
@@ -1,0 +1,44 @@
+# SPDX-FileCopyrightText: 2019 ash_postgres contributors <https://github.com/ash-project/ash_postgres/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
+defmodule AshPostgres.MigrationGenerator.OperationCycleError do
+  @moduledoc """
+  Raised when the migration operation dependency graph contains a cycle,
+  i.e. two or more operations require facts that only each other provide.
+  """
+  defexception [:operations]
+
+  @impl true
+  def message(%{operations: operations}) do
+    described =
+      operations
+      |> Enum.map(&describe_operation/1)
+      |> Enum.map(&"  - #{&1}")
+      |> Enum.join("\n")
+
+    """
+    Could not determine a valid order for the following migration operations \
+    because they form a dependency cycle:
+
+    #{described}
+
+    If two tables have foreign keys pointing at each other, split the \
+    change into multiple migrations (e.g. create both tables first, add \
+    the foreign keys afterward) or make one of them nullable/deferrable.
+
+    Otherwise, this is most likely a bug in AshPostgres's migration \
+    ordering logic rather than something you can fix in your resources \
+    — please open an issue at \
+    https://github.com/ash-project/ash_postgres/issues with the \
+    resources involved.
+    """
+  end
+
+  defp describe_operation(op) do
+    table = Map.get(op, :table)
+    schema = Map.get(op, :schema)
+
+    "#{inspect(op.__struct__)} on #{inspect(table)}#{if schema, do: " (schema: #{inspect(schema)})", else: ""}"
+  end
+end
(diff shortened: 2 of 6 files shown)


lib/migration_generator/operation.ex
```diff
@@ -157,13 +157,17 @@ defmodule AshPostgres.MigrationGenerator.Operation do
 
   defmodule RenameTable do
     @moduledoc false
-    defstruct [:old_table, :new_table, :schema, :multitenancy, :repo, no_phase: true]
+    # `:table` is the table's identity going forward (its post-rename name),
```
zachdaniel: I'm general comments should speak to how things are, not how they were.

lib/migration_generator/operation_cycle_error.ex
```diff
@@ -0,0 +1,43 @@
...
+      |> Enum.map(&"  - #{&1}")
+      |> Enum.join("\n")
+
+    """
+    Could not determine a valid order for the following migration operations \
+    because they form a dependency cycle:
+
+    #{described}
+
+    This usually means a structural foreign key (or other declared \
```
zachdaniel: If this happens it's not something the user can really do anything about or should have to address. We should explain that this is almost certainly a bug and that they should open an issue.

lib/migration_generator/operation_deps.ex
```diff
@@ -0,0 +1,340 @@
...
+        schema: schema,
+        new_attribute: new_attribute,
+        old_attribute: old_attribute
+      } ->
+        [
+          {:column_ready, key(table, schema, new_attribute.source)},
+          {:column_renamed_from, key(table, schema, old_attribute.source)},
+          {:any_attribute_added, key(table, schema)},
+          {:table_touched, key(table, schema)},
+          {:table_finalized, key(table, schema)}
```
zachdaniel: Thinking maybe we should describe these facts somewhere because I'm not fully understanding for example why adding an attribute provides "table finalized".

zachdaniel (commented):
This is very cool and I was always aware that the pairwise "after" was a hack, glad that you've invested the energy to create a better way 🤩

Have some comments/questions and will need to do a deeper review soon of the specific fact emissions etc.

lib/migration_generator/operation.ex
jechol: Good point — on reflection the comment wasn't adding anything the field names don't already say, so I just removed it instead of rewording it.

lib/migration_generator/operation_cycle_error.ex
jechol: Fixed — the message now says it's most likely a bug and asks to open an issue.

lib/migration_generator/operation_deps.ex
jechol: Added a "Facts" section to the moduledoc since — every fact is described there now. On the specific example: requiring a fact doesn't mean waiting for one provider — it creates an edge from every operation in the batch that provides it, and if nothing provides it, it's vacuously satisfied (the structure already exists from earlier migrations). That's why every structural op, including AddAttribute, provides the catch-all fact (since renamed to table_structure_ready): a single requirement then means "after all of this table's structural work in this batch" — or nothing at all if the table isn't being touched.

jechol:
Addressed all review comments (replies inline). The branch has changed quite a bit since your review — easier to re-review fresh than against the old diff. Two pre-existing ordering bugs were also found and fixed along the way (details in code comments and tests).

One removal worth flagging: the custom-statement guard in `streamline/2` (from 12cf97d8) is gone — it guarded against `op.schema` raising on structs that had no `:schema` field, which they now have, and the toposort can't place a same-table custom statement inside an add/alter window anyway.

Ready for another look.

jechol:
Ready for another pass — some context on what's changed since the last review, and why I think the dependency-graph model keeps paying off.

**New ordering facts found by real-world migrations since the last review**

Beyond translating the existing pairwise rules into the graph, running our production app's migrations against this branch kept surfacing orderings the pairwise model couldn't express. Each landed as a precise fact with a regression test:

1. **A foreign key must run after the unique index covering its referenced columns** (6431ffbe). Found while converting hand-written composite-FK custom statements to declarative `match_with` references: the FK was emitted before the `(org_id, id)` unique custom index it targets, failing with *"there is no unique constraint matching given keys"*. Since Postgres requires one unique index on **exactly** the referenced column set, this is modeled as a single column-set fact (`unique_index_created`) — built from the columns the rendered index actually covers (including the tenant attribute that attribute multitenancy prefixes at render time) and required for `destination_attribute` plus all `match_with` destinations. Two separate single-column unique indexes correctly do **not** satisfy a composite FK.
2. **Custom indexes with plain field lists now wait only on their own columns** (same commit). Previously a custom index waited for *every* column operation on its table. Combined with fact 1 that's a deadlock on the same table: the FK alter must wait for the unique index, but the unique index waited for the whole table to settle — which includes that very FK alter. Now an index whose fields are plain columns depends only on those columns; raw `where`/expression indexes keep the conservative whole-table wait.
3. **Rewriting a reference index must drop the old index before creating the new one** (1f4148c0), since both derive the same auto-generated name — and **a reference changing to `index? false` must not emit a fresh `AddReferenceIndex`** (50acf607). Both found the same way.

Locally everything is green: the full suite plus the `OperationDeps` unit tests covering each fact above.

zachdaniel:

PR #799, opened 2026-07-18
feat: add after_tables to custom_statements so raw SQL can depend on another table

jechol:
## Motivation

`custom_statements` had no way to declare "this depends on another table's structure being finalized first." The only rule was "adds always run last, removes always run first." That breaks for a raw-SQL statement that needs to run after some other table is fully set up — e.g. a composite foreign key referencing a unique index that isn't guaranteed to exist yet.

## What changed

`AshPostgres.Statement` gets a new `after_tables` option (a list of table name strings) so a statement can declare that dependency explicitly:

    custom_statements do
      statement :children_parent_composite_fk do
        after_tables ["parents"]
        up "ALTER TABLE children ADD CONSTRAINT children_parent_fk FOREIGN KEY (region_id, parent_id) REFERENCES parents (region_id, id);"
        down "ALTER TABLE children DROP CONSTRAINT children_parent_fk;"
      end
    end

Internally, #798 has a single per-table "doneness" fact, `:table_structure_ready` — provided by every structural operation (columns, indexes, constraints, etc.) on the table. This PR adds a second, broader fact on top of it:

- `:table_structure_ready` (existing) — this table's structural operations are done.
- `:table_finalized` (new) — the table is *truly* done, including any `custom_statements` declared on it. Provided by everything that provides `:table_structure_ready`, plus each `AddCustomStatement` on the table — so a table with no custom statements is finalized as soon as its structure is ready.

A statement's own implicit "wait for my own table" requirement uses the narrower `:table_structure_ready`, so two statements on the same table never end up requiring each other's completion (a guaranteed cycle if both used the broader fact). A declared `after_tables` entry uses the broader `:table_finalized`, so it also waits for the referenced table's own custom statements — not just its structure.

Statements with no `after_tables` keep the previous default behavior exactly (no declared dependency → best-effort last), so this is fully backwards compatible.

## Testing

Builds on the dependency-graph rewrite in #798. New tests in `test/migration_generator/operation_deps_test.exs` cover: a statement's own-table requirement being satisfied by structural ops, an `after_tables` requirement being satisfied by another table's custom statement (the actual new capability), and two same-table statements not creating a cycle. Full `mix test`: 858 passed.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:
Ready for this to be backmerged

jechol:
Done — rebased onto `main`, full suite passes (879 tests). @zachdaniel ready for another look.

zachdaniel:

PR #800, opened 2026-07-21
fix: fix generated reference type value for identity and bigserial

vonagam:
Don't know how it worked before but all `reference_type` calls are wrapped in `inspect` and if the value is `":bigserial"` instead of `:bigserial` then you end up with `type: ":bigserial"` in a generated migration and for that Ecto will throw an unsupported type error...


zachdaniel (approved):

zachdaniel:
🤔 weird

PR #806, opened 2026-07-26
fix: clean up obsolete serial sequences

jechol:
Changing a generated ID from an implicit `bigserial` default to an explicit default leaves its owned sequence behind. This causes migrated schemas to differ from freshly generated schemas.

This change removes obsolete sequences in the correct order and restores them in reverse migrations. Restored sequences are initialized from the column's current maximum value.

Regression tests cover both transition directions. I also verified that migrated and freshly generated schemas match in a downstream application.

This pull request description was drafted with AI assistance and reviewed by the submitter.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel (approved):
clever

PR #809, opened 2026-07-29
test: cover bulk upsert dropping records when select excludes identity keys

marcnnn:
On upsert the columns that need to be selected to match the updated rows with the rows from the database are not selected anymore so upsert result is empty even if the data is written into the db.

This is a regression that is related to https://github.com/ash-project/ash_postgres/pull/786 and https://github.com/ash-project/ash_postgres/pull/638

In my case, an upsert_identity with unique_index defines the upsert semantic already, so the select was left out and is filled by the upsert reactor with the empty list. But the unique index is constructed with two fields not only one like the test, i can add this as well?

My dependabot branch for the ash_postgres update from 2.10.0 to 2.11.0 failed CI
The tests reproduce a simplified version of the problem.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/test/bulk_create_test.exs
+++ b/test/bulk_create_test.exs
@@ -323,6 +323,64 @@ defmodule AshPostgres.BulkCreateTest do
                end)
     end
 
+    # Bulk upserts correlate the returned rows back to their changesets by the upsert
+    # identity's keys, so those keys have to be read back even when the action's select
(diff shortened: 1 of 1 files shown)


zachdaniel:
tests added in a separate PR, thanks!

Issue #243, opened 2024-04-22
Incorrect validation error when using AshPostgres


albinkc:
* creating new issue in case the original comment flies under the radar *

I'm getting a validation error, even though the status in the create response and DB says open (verified manually).

```elixir
  ticket = (  Helpdesk.Support.Ticket  |> Ash.Changeset.for_create(:open, %{subject: "My mouse won't click33!"})  |> Ash.create!())

22:09:03.164 [debug] QUERY OK db=2.4ms idle=1348.9ms
begin []

22:09:03.199 [debug] QUERY OK source="tickets" db=3.0ms
INSERT INTO "tickets" ("id","status","subject") VALUES ($1,$2,$3) RETURNING "status","subject","id" ["b201e6eb-f752-4f0d-b1d3-32013c47f8d0", :open, "My mouse won't click33!"]

22:09:03.204 [debug] QUERY OK db=2.1ms
commit []
#Helpdesk.Support.Ticket<
  __meta__: #Ecto.Schema.Metadata<:loaded, "tickets">,
  id: "b201e6eb-f752-4f0d-b1d3-32013c47f8d0",
  subject: "My mouse won't click33!",
  status: :open,
  aggregates: %{},
  calculations: %{},
  ...
>
iex(2)> ticket|> Ash.Changeset.for_update(:close)|> Ash.update!()
** (Ash.Error.Invalid) Invalid Error

* Invalid value provided for status: Ticket is already closed.

"closed"

  (elixir 1.16.2) lib/process.ex:860: Process.info/2
  (ash 3.0.0-rc.27) lib/ash/error/changes/invalid_attribute.ex:5: Ash.Error.Changes.InvalidAttribute.exception/1
  (ash 3.0.0-rc.27) lib/ash/query/function/error.ex:13: Ash.Query.Function.Error.evaluate/1
  (ash 3.0.0-rc.27) lib/ash/filter/runtime.ex:578: Ash.Filter.Runtime.resolve_expr/5
  (ash 3.0.0-rc.27) lib/ash/filter/runtime.ex:231: Ash.Filter.Runtime.do_match/5
  (ash 3.0.0-rc.27) lib/ash/changeset/changeset.ex:2203: anonymous fn/4 in Ash.Changeset.add_atomic_validations/3
  (elixir 1.16.2) lib/enum.ex:4839: Enumerable.List.reduce/3
  (elixir 1.16.2) lib/enum.ex:2582: Enum.reduce_while/3
[...]

PR #811, opened 2026-08-03
chore: add tests for ash_sql #243

matt-beanland:
# Contributor checklist

- [X] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

Reproduction/regression test for: https://github.com/ash-project/ash_sql/pull/244 (issue:
https://github.com/ash-project/ash_sql/issues/243).

`Ash.Query.combination_of/2` applies its parts in the order given, each to the result of
everything before it. `ash_sql` renders them as a flat chain of set operations, so SQL's own
precedence regroups a combination that spans more than one precedence level .
INTERSECT` binds tighter than `UNION`/`EXCEPT`. The query then answers differently here than it does under `Ash.DataLayer.Ets`.

The test adds a three-part case to `test/combination_test.exs`: `base`, `union`, then `intersect`,
where applying the parts in order and letting SQL regroup them give different answers. A fourth
record belongs to neither grouping, so reading the whole table cannot be mistaken for a pass.

```
against ash_sql 0.6.6      left:  ["alpha", "beta", "gamma"]
                           right: ["alpha", "gamma"]

with ash_sql#NNN           19 passed
```

Every existing combination test here stops at two parts, which is the one shape where the order
cannot be observed — two parts agree however they are grouped, and `union`/`except` share a
precedence in SQL so they agree too. That is why this has not surfaced.

⚠️ This test fails until ash_sql#244 lands and is released. Note also that `main` is currently red
independently of this: 11 tests in `AshPostgres.AggregateReadActionTest` fail on `fc165f3e`
against released `ash_sql` 0.6.6, unrelated to combinations or to this change.

### DIFF
--- a/test/combination_test.exs
+++ b/test/combination_test.exs
@@ -222,6 +222,34 @@ defmodule AshPostgres.CombinationTest do
       assert hd(result).title == "post1"
     end
 
+    test "a third combination part applies to the running result, not to the second part" do
+      for title <- ["alpha", "beta", "gamma", "delta"] do
+        Post
+        |> Ash.Changeset.for_create(:create, %{title: title})
+        |> Ash.create!()
+      end
+
+      # The parts are applied in the order given, so this is
+      # ({alpha, beta} ∪ {gamma}) ∩ {alpha, gamma}.
+      #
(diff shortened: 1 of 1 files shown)


zachdaniel:

PR #808, opened 2026-07-27
fix: return upsert identity keys so bulk upserts don't drop records

marcnnn:
My dependabot branch for the ash_postgres update from 2.10.0 to 2.11.0 failed CI. 

I send my AI to create a fix.

I will read the surrunding code until I can verify the AI code but open this as draft until then.
# Contributor checklist

Leave anything that you believe does not apply unchecked.

- [x] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [x] Bug fixes include regression tests
- [ ] Chores
- [ ] Documentation changes
- [ ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

PR #812, opened 2026-08-11
fix: return upsert identity keys and correlate on the full key set so bulk upserts don't drop
    records

DGollings:
## Problem

Since #786, bulk upserts can silently drop records from the returned result while still
writing them to the database — so `after_action` hooks and notifiers never run for the
dropped records. This surfaced on the 2.10.0 → 2.11.0 upgrade (same regression as #808/#809).

`bulk_create/3` correlates each returned row back to its changeset by the upsert identity's
keys. Two things break that correlation:

1. **The identity keys aren't selected.** When the action's `select` excludes them
   (e.g. `upsert_fields []` with `select: []`, as `Ash.Reactor`'s `bulk_create` step passes),
   the returned rows don't contain the identity columns, so nothing correlates and every
   record is dropped. This is the fix from @marcnnn's #808, with the regression tests from #809.

2. **Unset (nil) identity keys are missing on the changeset side.** For a multi-field identity
   that leaves a key unset — a nullable field left `nil`, common with `nils_distinct?: false`
   natural keys such as `[barcode, timestamp, order_id, route_id]` — the changeset key is built
   with `Map.take(changeset.attributes, keys)`, which drops the unset key. The changeset's
   correlation key then has fewer fields than the returned row's key (whose `Map.take(row, keys)`
   includes the key as `nil`), so it still never matches even after (1).

## Fix

- Add the identity's attribute keys to the upsert `RETURNING` list so returned rows always
  carry them. `Ash.Actions.Helpers.select/2` masks the extra fields back out of the records,
  so this doesn't widen what callers observe. Identity keys that aren't attributes (e.g. an
  identity over a calculation) have no column and are left out.
- Build the per-changeset correlation key over the full identity key set, defaulting unset
  keys to `nil`, so both sides have the same shape.

## Tests

- `bulk upsert returns inserted records when select excludes the identity keys` (from #809)
- `bulk upsert returns updated records when select excludes the identity keys` (from #809)
- `bulk upsert returns records when a multi-field identity leaves a key unset (nil)` (new;
  covers facet 2)

The fix in (1) and the first two tests are originally by @marcnnn (#808, #809); this PR carries
them and adds the multi-field / nil-identity correlation fix and test on top.

## Note for reviewers

On PostgreSQL 17 (MERGE path) the pre-existing test `returns a skipped upsert whose identity
contains nil` (added by #794) fails with a unique-constraint error — that's a separate
nil-identity issue in the `return_skipped_upsert?` path, unrelated to and not addressed by
this PR.

## Disclosure

Drafted with AI assistance, but every change was reviewed and validated: the failure is
reproduced by a failing test first, the fix is verified against those tests, and the whole
change was exercised end-to-end against a real 8k-test downstream suite on PostgreSQL 17.

---

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -2443,7 +2443,7 @@ defmodule AshPostgres.DataLayer do
             |> Enum.filter(fn changeset ->
               not Map.has_key?(
                 results_by_identity,
-                Map.take(changeset.attributes, keys)
+                changeset_correlation_key(changeset, keys)
               )
             end)
             |> Enum.map(fn changeset ->
@@ -2483,9 +2483,7 @@ defmodule AshPostgres.DataLayer do
           results =
             changesets
             |> Enum.map(fn changeset ->
-              identity =
-                changeset.attributes
-                |> Map.take(keys)
+              identity = changeset_correlation_key(changeset, keys)
 
               Map.get(results_by_identity, identity, Map.get(skipped_upserts, identity))
             end)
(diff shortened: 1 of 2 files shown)


zachdaniel:

Issue #553, opened 2025-05-22
support for Duration


matt-beanland:
https://github.com/ash-project/ash/pull/2036 is adding Ash.Type.Duration (short name :duration), ash_postgres should store Duration structure as Postgres interval, and should handle expressions with operators involving durations in the datalayer.

zachdaniel:
Just needs to be using latest versions of `ash_sql` and `ash` for those tests to pass I believe.

zachdaniel:
The tests are very slow, try skipping migration generator tests with `mix test --exclude migration`

zachdaniel:
I guess you said you updated deps though, so maybe something is wrong.

zachdaniel:
Would that mean the end user can't use the built in duration type? Ideally we'd wrap that in ash_postgres automatically so they can just use durations and we just handle that when type casting automatically.

matt-beanland:
@zachdaniel this shouldn't have been closed yet. Can you please re-open?

I've raised https://github.com/ash-project/ash/issues/2850 and will contribute to PR.
There is then work required on the mapping to/from PG Interval which is trivial, and work on the operators . PR's likely to ash, ash_sql and ash_postgres.

matt-beanland:
The two outstanding aspects of this issue are now PR's:

persistence - #815 

operators - ash#2852 (fix) and #816 (tests). 

#816 will be red until ash#2852 releases, the same way #814 was until ash_sql 0.6.8.

Worth noting for ordering: #815 is best merged after ash#2851, otherwise a `units`-constrained duration becomes persistable while still not round-tripping.

matt-beanland:
Closing as these were all done

PR #815, opened 2026-08-15
improvement: decode an interval as a Duration by default

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Part of #553 (doesn't close it because operators are outstanding)

A `:duration` attribute cannot be read on Postgres. The value encodes into the `interval` column fine, but the `RETURNING` row loads back as a `%Postgrex.Interval{}`, which `Ash.Type.Duration` refuses, then the create fails with an `Ash.Error.Unknown` and the insert is rolled back, leaving nothing stored.

postgrex can already decode an interval as a `Duration`, but `interval_decode_type` is only settable at `Postgrex.Types.define/3`, never in repo configuration, and ash_postgres shipped no types module. This adds `AshPostgres.PostgrexTypes` and applies it in `AshPostgres.Repo.init/2` with `Keyword.put_new`, so a repo defining its own module for other extensions keeps it.

The migration generator needs no change: a `:duration` attribute already generates `add(:field, :duration)`, which Ecto renders as `interval`. In Postgres `INTERVAL` and `INTERVAL(6)` are identical with microsecond precision.

This changes interval decoding for every repo, not only `:duration` attributes: raw queries and any custom interval-mapped type will receive a `Duration` rather than a `%Postgrex.Interval{}`. There is no Ash `:duration` data to migrate, since loading one currently fails.

Best paired with ash#2851 — without it, a `units`-constrained duration is readable but not re-writable once persisted.

Tested with Postgres 19.

### DIFF
--- a/test/support/postgrex_types.ex
+++ b/test/support/postgrex_types.ex
@@ -1,0 +1,11 @@
+# SPDX-FileCopyrightText: 2019 ash_postgres contributors <https://github.com/ash-project/ash_postgres/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
+# The test repos configure this exactly as the `AshPostgres.Repo` docs tell users to,
+# so `interval_decode_type: Duration` is exercised rather than only described.
+Postgrex.Types.define(
+  AshPostgres.Test.PostgrexTypes,
+  Ecto.Adapters.Postgres.extensions(),
+  interval_decode_type: Duration
+)
--- a/config/config.exs
+++ b/config/config.exs
@@ -50,7 +50,8 @@ if Mix.env() == :test do
     username: "postgres",
     database: "ash_postgres_test",
     hostname: "localhost",
-    pool: Ecto.Adapters.SQL.Sandbox
+    pool: Ecto.Adapters.SQL.Sandbox,
+    types: AshPostgres.Test.PostgrexTypes
 
   config :ash_postgres, AshPostgres.DevTestRepo,
     username: "postgres",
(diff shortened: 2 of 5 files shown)


test/support/postgrex_types.ex
```diff
@@ -0,0 +1,11 @@
+# SPDX-FileCopyrightText: 2019 ash_postgres contributors <https://github.com/ash-project/ash_postgres/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
+# `interval_decode_type` can only be set at `Postgrex.Types.define/3`, never in repo
+# configuration, so an `interval` column needs this module to load as a `Duration`.
+Postgrex.Types.define(
```
zachdaniel, line 7: This could theoretically break any custom code that a user has. In other cases, we require the user to do this in their own application so that they have a single authoritative source of truth. I think we will have to do that here as well otherwise risk anyone who has custom interval handling code that is also using AshPostgres.

matt-beanland: Yes, agreed. The default is gone, and a repo's `types:` is never touched.

Now:

* The error message teaches - A `%Postgrex.Interval{}` reaching `Ash.Type.Duration` now appends
the `Postgrex.Types.define/3` call and the repo config to the message, instead of a cast
error naming neither.

* The tests show how - `AshPostgres.TestRepo` configures its own types module exactly as
the docs instruct, so the documented path is the one CI exercises rather than only
describes.

* The solution is pinned - Tests assert an `interval` loads as a `Duration`, that
months and days survive separately, and that a repo configuring its own `types:` keeps it.

(GitHub has followed this comment onto the test-support module — the file it was on is
deleted.)

PR #814, opened 2026-08-14
chore: add tests for the Duration form of the datetime expressions

matt-beanland:
# Contributor checklist

Leave anything that you believe does not apply unchecked.

- [X] I accept the [AI Policy](https://github.com/ash-project/.github/blob/main/AI_POLICY.md), or AI was not used in the creation of this PR.
- [ ] Bug fixes include regression tests
- [X] Chores
- [ ] Documentation changes
- [ ] Features include unit/acceptance tests
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Tests only. Covers `ago/1`, `from_now/1`, `datetime_add/2` and `date_add/2` with a `Duration`, including a multi-unit duration and a negative one, and checks that `date_add` still compares as a date.

These pass only with the corresponding ash_sql change; against released ash_sql they fail with `Unsupported expression`. Part of #553 

Tested with Postgres 19.

# Notes
requires https://github.com/ash-project/ash/pull/2844 for `from_now/1`
requires https://github.com/ash-project/ash_sql/pull/245

PR #816, opened 2026-08-16
test: cover the duration operators on a datetime attribute

matt-beanland:
# Contributor checklist

- [X] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

Partially delivers #553 (Duration)
Requires ash PR https://github.com/ash-project/ash/pull/2852 to **release** before this merges

Tests only. This PR tests `attribute + duration`, `attribute - duration`, a multi-unit duration, the duration on the left, and both operands as expressions.

Against released ash 3.31.3 all six fail with `Invalid filter value %Duration{...}`, because `plus`/`minus` declare no types and the duration is typed as the datetime it is added to. CI here resolves ash from Hex, so this stays red until that releases; as #814 was until ash_sql 0.6.8.

Independent of #815.

Tested with Postgres 19.

### DIFF
--- a/test/duration_operator_test.exs
+++ b/test/duration_operator_test.exs
@@ -1,0 +1,90 @@
+# SPDX-FileCopyrightText: 2019 ash_postgres contributors <https://github.com/ash-project/ash_postgres/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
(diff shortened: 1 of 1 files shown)


zachdaniel:
I'm cool with tests failing for now 😄

Issue #823, opened 2026-08-19
Valid custom-statement dependencies produce an artificial operation cycle


Alt-iOS:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

ash 3.31.3
ash_postgres 2.12.0
Erlang/OTP 29 [erts-17.0.3] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1] [jit]
Elixir 1.20.2 (compiled with Erlang/OTP 29)

### Operating system

macos 27 Beta 4

### Current Behavior

 `after_tables` waits for the referenced table to be fully finalized, including its custom statements. Valid cross-table dependencies can therefore produce a false `OperationCycleError`.

For example, a statement on table A requiring table B’s structure, and a statement on B requiring A’s structure, is treated as a statement cycle.

### Reproduction

  Define custom statements on two resources:
```elixir  
# Resource A
after_tables ["table_b"]
# Resource B
after_tables ["table_a"]
```

  Each statement should reference only the other table’s structure. Then run:
`mix ash_postgres.generate_migrations`

Generation fails with AshPostgres.MigrationGenerator.OperationCycleError, although both tables could be created before either statement. This is a full failure test
```elixir
test "orders cross-table and same-resource custom statement dependencies", %{
    snapshot_path: snapshot_path,
    migration_path: migration_path
  } do
    defresource DependencyParent do
      postgres do
        table "dependency_parents"
        repo AshPostgres.TestRepo

        custom_statements do
          statement :parent_view do
            # Proposed replacement:
            # after_table_structures ["dependency_children"]
            # after_statements [:parent_function]
            after_tables ["dependency_children"]

            up """
            CREATE VIEW dependency_parent_view AS
            SELECT parent_value(id) FROM dependency_children
            """

            down "DROP VIEW dependency_parent_view"
          end

          statement :parent_function do
            up """
            CREATE FUNCTION parent_value(value bigint)
            RETURNS bigint
            LANGUAGE SQL
            IMMUTABLE
            AS 'SELECT value'
            """

            down "DROP FUNCTION parent_value(bigint)"
          end
        end
      end

      attributes do
        attribute :id, :integer, primary_key?: true, allow_nil?: false
      end
    end

    defresource DependencyChild do
      postgres do
        table "dependency_children"
        repo AshPostgres.TestRepo

[...]

PR #826, opened 2026-08-20
Fix/custom statement dependencies

Alt-iOS:
This adds a custom_statement ordering for same resource custom statements (ie, a function before its trigger), makes `after_tables` mean only the table structure and renames `after_tables` to `after_resources` to better reflect that it means after the table and custom statements. Addresses #823 

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: MIT
 
 spark_locals_without_parens = [
+  after_statements: 1,
+  after_table_structures: 1,
   after_tables: 1,
   all_tenants?: 1,
   base_filter_sql: 1,
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -3,8 +3,8 @@
 # SPDX-License-Identifier: MIT
 
 spark_locals_without_parens = [
+  after_resource: 1,
   after_statements: 1,
-  after_table_structures: 1,
   after_tables: 1,
   all_tenants?: 1,
   base_filter_sql: 1,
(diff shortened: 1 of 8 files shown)


zachdaniel:
For ordering within statements, I think we could simply use the order that they appear in the resource. For `after_resource`, I think that is a confusingly named thing because multiple resources can share the same table. I'm still not fully convinced that we need it if `after_tables` is repurposed to just "after the table structure".

Alt-iOS:
For the after_statements part, let me see if I can make it deterministic. The migration generator stopped preserving declaration order once one of the `custom_statements` had an `after_tables` dependency. I’ll see if I can preserve declaration order correctly in that case, which should remove the need for `after_statements`.

As for `after_resource`, the name is certainly a choice. I only suggested keeping the old behavior under another name because it was explicitly created for that purpose in the pr that added it. But it's only been in version 2.12.0 so I doubt many are using it, so f you don’t think that behavior needs to remain supported, I’ll remove it and clean up the old facts, comments, and tests in the next commit.

Alt-iOS:
@zachdaniel Pushed deleting the `after_statements` and `after_resource`, now this
```elixir
custom_statements do
  statement :create_function do
    after_tables ["audit_entries"]
    up "CREATE FUNCTION ..."
    down "DROP FUNCTION ..."
  end

  statement :create_trigger do
    up "CREATE TRIGGER ... EXECUTE FUNCTION ..."
    down "DROP TRIGGER ..."
  end
end
```
Creates this migration order
```
create widgets table structure
create audit_entries table structure
create function
create trigger
```

zachdaniel:

PR #836, opened 2026-08-21
fix: run tenant migrations from mix ash.setup

### LINKED ISSUE #807: `mix ash.setup` never applies tenant migrations: inverted `has_tenant_migrations?` predicate, and the second `ash_postgres.migrate` invocation is a silent no-op
## Summary

`AshPostgres.DataLayer.setup/1` (what `mix ash.setup` delegates to) is supposed to run the tenant migration pass after the regular one, but three compounding defects mean the tenant pass can **never** actually apply tenant migrations, in any configuration:

1. **Inverted predicate** — `has_tenant_migrations?/1` returns `Enum.empty?/1`, i.e. `true` when there are *no* tenant migrations. Combined with the `Enum.all?(&(not has_tenant_migrations?(&1)))` guard, `setup/1` skips the tenant pass whenever every configured repo has tenant migrations — for the common single-repo case, exactly when tenant migrations *do* exist — and only attempts it when there is nothing to migrate.
2. **Mix per-name task dedup** — even when the tenant branch is reached, `Mix.Task.run("ash_postgres.migrate", ...)` is a silent no-op: the same task already ran a few lines above, Mix deduplicates tasks *by name* (arguments are not part of the key), and `ash_postgres.migrate` never re-enables itself (it only re-enables `ecto.migrate` internally). The same trap affects the `test.reset` alias in this repo's own `mix.exs`.
3. **Invalid flag (latent)** — the tenant branch passes `--tenant` (singular), but the task's strict `OptionParser` switches only define `tenants: :boolean`. If defect 2 were fixed in isolation, this line would raise `OptionParser.ParseError` instead of running the tenant pass.

Verified on current `main` (d059a4494b183a02ee4d2b56fdc7b70e8e804a40, v2.11.0 in `mix.exs`); we originally hit it on 2.6.32.

### 1. `has_tenant_migrations?` is inverted

https://github.com/ash-project/ash_postgres/blob/d059a4494b183a02ee4d2b56fdc7b70e8e804a40/lib/data_layer.ex#L657-L663

Quoted below with an inline annotation of ours:

```elixir
defp has_tenant_migrations?(repo) do
  []
  |> AshPostgres.Mix.Helpers.tenant_migrations_path(repo)
  |> Path.join("**/*.exs")
  |> Path.wildcard()
  |> Enum.empty?()          # <- returns true when there are NO tenant migrations
end
```

Used here:

https://github.com/ash-project/ash_postgres/blob/d059a4494b183a02ee4d2b56fdc7b70e8e804a40/lib/data_layer.ex#L635-L650

```elixir
def setup(args) do
  # TODO: take args that we care about
  Mix.Task.run("ash_postgres.create", args)
  Mix.Task.run("ash_postgres.migrate", args)

  []
  |> AshPostgres.Mix.Helpers.repos!(args)
  |> Enum.all?(&(not has_tenant_migrations?(&1)))
  |> case do
    true ->
      :ok

    _ ->
      Mix.Task.run("ash_postgres.migrate", ["--tenant" | args])
  end
end
```

Reading the guard with the inverted predicate: `Enum.all?(&(not Enum.empty?(wildcard)))` is `true` when **every repo has tenant migrations** — and that is the branch that skips (`:ok`). So the common case (a multitenant app with tenant migrations) never reaches the tenant pass. The `Enum.all?` guard itself is written for the correct semantics; only the predicate body is inverted.

### 2. The second `ash_postgres.migrate` invocation is a no-op

Even in the branch where line 648 runs, nothing happens. `Mix.Task.run/2` runs each task **once per name** per Mix invocation:

> If the task or alias has already been invoked, subsequent calls to `run/2` will abort without executing and return `:noop`. (Elixir docs, `Mix.Task.run/2`)

The dedup key in `Mix.TasksServer` is `{:task, task_name, project}` — arguments are not part of it, so `ash_postgres.migrate` and `ash_postgres.migrate --tenants` are the same task in Mix's eyes. Since `Mix.Tasks.AshPostgres.Migrate.run/1` only re-enables `ecto.migrate` (never itself), the first invocation at line 638 consumes the task and the second at line 648 returns `:noop` with no output.

Minimal repro of the Mix semantics (toy project, task `demo.say` that prints its args):

```elixir
demo_broken: ["demo.say alpha", "demo.say beta"]
# => demo.say executed with args: ["alpha"]        (beta is NEVER executed)

demo_fixed: ["demo.say alpha", fn _ -> Mix.Task.rerun("demo.say", ["beta"]) end]
# => demo.say executed with args: ["alpha"]
# => demo.say executed with args: ["beta"]
```

The same trap exists in this repo's own aliases:

https://github.com/ash-project/ash_postgres/blob/d059a4494b183a02ee4d2b56fdc7b70e8e804a40/mix.exs#L265

```elixir
"test.reset": ["test.drop", "test.create", "test.migrate", "ash_postgres.migrate --tenants"],
```

`test.migrate` is itself an alias for `ash_postgres.migrate`, so by the time the last item runs, the task name is already marked and the `--tenants` pass silently no-ops.

### 3. `--tenant` is not a valid flag

Line 648 passes `["--tenant" | args]`, but the migrate task parses strictly and only knows the plural form:

https://github.com/ash-project/ash_postgres/blob/d059a4494b183a02ee4d2b56fdc7b70e8e804a40/lib/mix/tasks/ash_postgres.migrate.ex#L18-L36 (`tenants: :boolean`)
https://github.com/ash-project/ash_postgres/blob/d059a4494b183a02ee4d2b56fdc7b70e8e804a40/lib/mix/tasks/ash_postgres.migrate.ex#L113-L115 (`OptionParser.parse!(args, strict: @switches, ...)`)

Today this is masked by defect 2 (the call never really executes). Fixing the no-op alone would turn the silent skip into an `OptionParser.ParseError` crash.

## Impact / how we hit this

On a project with schema-based multitenancy and a **persistent** dev/test database, tenant schemas silently drift: `mix ash.setup` reports success, all public-schema migrations are applied, but new tenant migrations are never run. We chased `undefined_column` errors on tenant tables before finding that the tenant pass had never executed and running `mix ash_postgres.migrate --tenants` by hand.

The bug is well hidden because it is often invisible on a freshly seeded database when tenants are created through `create_tenant!/2`: `AshPostgres.MultiTenancy.create_tenant!/2` auto-migrates each tenant schema at creation time (https://github.com/ash-project/ash_postgres/blob/d059a4494b183a02ee4d2b56fdc7b70e8e804a40/lib/multitenancy.ex#L11-L18), so seeds that create tenants leave them fully migrated. The gap only opens when new tenant migrations land *after* the tenant schemas already exist — at which point `mix ash.setup` does not apply them.

## Suggested fix

All three are small:

```elixir
def setup(args) do
  # TODO: take args that we care about
  Mix.Task.run("ash_postgres.create", args)
  Mix.Task.run("ash_postgres.migrate", args)

  []
  |> AshPostgres.Mix.Helpers.repos!(args)
  |> Enum.any?(&has_tenant_migrations?(&1))
  |> case do
    false ->
      :ok

    _ ->
      Mix.Task.rerun("ash_postgres.migrate", ["--tenants" | args])
  end
end

defp has_tenant_migrations?(repo) do
  # ...
  |> Path.wildcard()
  |> Enum.empty?()
  |> Kernel.not()
end
```

- Fix the predicate so it matches its name (`not Enum.empty?`).
- Use `Mix.Task.rerun/2` for the second invocation — or the equivalent explicit `reenable` + `run` pair this file already uses in `migrate/1`: https://github.com/ash-project/ash_postgres/blob/d059a4494b183a02ee4d2b56fdc7b70e8e804a40/lib/data_layer.ex#L440-L443
- `--tenant` → `--tenants`.
- Same `rerun` treatment for the last item of the `test.reset` alias in `mix.exs` (Mix aliases accept 1-arity functions as items): `fn args -> Mix.Task.rerun("ash_postgres.migrate", ["--tenants" | args]) end`.

Happy to send a PR if this direction looks good to you.

TravelCurry02:
Fixes #807.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

`mix ash.setup` was supposed to run tenant migrations after public ones, but it never did.

Three bugs:
- `has_tenant_migrations?/1` returned true when there were *no* tenant files
- the second `ash_postgres.migrate` call was skipped because Mix won't run the same task twice (`run` vs `rerun`)
- it passed `--tenant` instead of `--tenants`

This inverts the check, uses `Mix.Task.rerun/2` with `--tenants`, does the same for the `test.reset` alias, and adds regression tests.

### DIFF
--- a/lib/data_layer.ex
+++ b/lib/data_layer.ex
@@ -639,13 +639,13 @@ defmodule AshPostgres.DataLayer do
 
     []
     |> AshPostgres.Mix.Helpers.repos!(args)
-    |> Enum.all?(&(not has_tenant_migrations?(&1)))
+    |> Enum.any?(&has_tenant_migrations?(&1))
     |> case do
-      true ->
+      false ->
         :ok
 
       _ ->
-        Mix.Task.run("ash_postgres.migrate", ["--tenant" | args])
+        Mix.Task.rerun("ash_postgres.migrate", ["--tenants" | args])
     end
   end
 
@@ -660,6 +660,7 @@ defmodule AshPostgres.DataLayer do
     |> Path.join("**/*.exs")
     |> Path.wildcard()
     |> Enum.empty?()
+    |> Kernel.not()
   end
 
   import Ecto.Query, only: [from: 2, subquery: 1]
--- a/mix.exs
+++ b/mix.exs
@@ -262,7 +262,12 @@ defmodule AshPostgres.MixProject do
       "test.migrate": "ash_postgres.migrate",
       "test.rollback": "ash_postgres.rollback",
       "test.create": "ash_postgres.create",
-      "test.reset": ["test.drop", "test.create", "test.migrate", "ash_postgres.migrate --tenants"],
+      "test.reset": [
+        "test.drop",
+        "test.create",
+        "test.migrate",
+        fn args -> Mix.Task.rerun("ash_postgres.migrate", ["--tenants" | args]) end
+      ],
       "test.drop": "ash_postgres.drop"
     ]
   end
--- a/test/data_layer_setup_test.exs
+++ b/test/data_layer_setup_test.exs
@@ -1,0 +1,64 @@
+# SPDX-FileCopyrightText: 2019 ash_postgres contributors <https://github.com/ash-project/ash_postgres/graphs/contributors>
+#
+# SPDX-License-Identifier: MIT
+
+defmodule Mix.Tasks.AshPostgres.SetupProbe do
+  @moduledoc false
+  use Mix.Task
+
+  @impl true
+  def run(args) do
+    send(:ash_postgres_setup_probe, {:ash_postgres_setup_probe, args})
+    :ok
+  end
+end
+
+defmodule AshPostgres.DataLayerSetupTest do
+  @moduledoc false
+  use ExUnit.Case, async: false
+
+  describe "has_tenant_migrations?/1" do
+    test "TestRepo has tenant migration files, so setup takes the tenant pass" do
+      files =
+        []
+        |> AshPostgres.Mix.Helpers.tenant_migrations_path(AshPostgres.TestRepo)
+        |> Path.join("**/*.exs")
+        |> Path.wildcard()
+
+      refute Enum.empty?(files)
+    end
+  end
+
+  describe "ash_postgres.migrate switches" do
+    test "rejects --tenant (singular), which is what setup used to pass" do
+      assert_raise Mix.Error, fn ->
+        Mix.Task.rerun("ash_postgres.migrate", ["--tenant"])
+      end
+    end
+  end
+
+  describe "Mix.Task.run/2 vs Mix.Task.rerun/2" do
+    setup do
+      Process.register(self(), :ash_postgres_setup_probe)
+      Mix.Task.reenable("ash_postgres.setup_probe")
+
+      on_exit(fn ->
+        if Process.whereis(:ash_postgres_setup_probe) == self() do
+          Process.unregister(:ash_postgres_setup_probe)
+        end
+      end)
+
+      :ok
+    end
+
+    test "a second Mix.Task.run of the same task is a no-op; Mix.Task.rerun is not" do
+      Mix.Task.run("ash_postgres.setup_probe", ["first"])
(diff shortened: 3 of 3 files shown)


zachdaniel:

PR #843, opened 2026-09-02
test: cover multi-hop exists aggregates and exists/2 over relationship paths

barnabasJ:
Companion to ash-project/ash_sql#259 — regression tests for multi-hop `exists` wrongly returning `true` with zero related rows when the predicate has no refs (unfiltered `exists` aggregates, `exists(path, true)`).

## What's added

New self-contained resources `AshPostgres.Test.TwoHopExistsTest.{Entry, Bucket, Item}` (Entry `belongs_to` Bucket, Bucket `has_many` Items, plus an `entry` belongs_to and a `no_attributes?` relationship on Item), with tests appended to the existing `describe "exists"` block in `test/aggregate_test.exs`:

- multi-hop exists aggregate + `exists/2` are **false with zero related rows** (the original bug; also asserts the tail table appears in the generated SQL), and true with one related row
- three-hop path with the chain broken at each level
- a broken middle hop is not masked by a `no_attributes?` (`on: true`) hop later in the path — this one requires anchoring *every* hop, not just the last
- exists predicates bind to the **last** resource in the path (all three tables share a `name` column; every row except the leaf is named to bait a wrong binding)
- single-hop baseline

## Note on CI

These tests fail against ash_sql 0.7.1 from hex (that's the bug) and pass with ash-project/ash_sql#259 (`ASH_SQL_VERSION=local`), where the full suite is green including all pre-existing exists/aggregate tests. This PR should land after/with the ash_sql fix.


zachdaniel:

Issue #805, opened 2026-07-23
Composite foreign keys result in incorrect migration


LostKobrakai:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### AI Policy

- [x] I agree to follow this project's AI Policy, or I agree that AI was not used while creating this issue.

### Versions

```
Erlang/OTP 29 [erts-17.0.2] [source] [64-bit] [smp:14:14] [ds:14:14:10] [async-threads:1] [jit]

Elixir 1.20.2 (compiled with Erlang/OTP 27)
```

```
* ash 3.30.1 (Hex package) (mix)
  locked at 3.30.1 (ash) a30153e1
* ash_postgres 2.11.0 (Hex package) (mix)
  locked at 2.11.0 (ash_postgres) c96d2776
```

### Operating system

Mac OS

### Current Behavior

A composite (`match_with`) foreign key gets wired up via `alter ... modify` before the target table's own match_with column exists. That match_with column only gets added with alter table later.

### Reproduction

```elixir
Mix.install([
  {:ash, "~> 3.0"},
  {:ash_postgres, "2.11.0"}
])

# NOTE: this appears sensitive to naming/ordering tie-breaks in the
# generator's dependency resolution - renaming these resources (e.g. to
# Tenant/Parent/Child) made the bug disappear in testing, so these exact
# names are load-bearing for reproducing it.

defmodule Repro.Repo do
  use AshPostgres.Repo, otp_app: :repro
end

Application.put_env(:repro, Repro.Repo, [])

defmodule Repro.Site do
  use Ash.Resource, domain: nil, data_layer: AshPostgres.DataLayer

  actions do
    defaults([:create, :read, :update, :destroy])
  end

  attributes do
    uuid_primary_key(:id)
  end

  postgres do
    table "sites"
    repo Repro.Repo
  end
end

defmodule Repro.A do
  use Ash.Resource, domain: nil, data_layer: AshPostgres.DataLayer

  actions do
    defaults([:create, :read, :update, :destroy])
  end

  attributes do
    uuid_primary_key(:id)
  end

  relationships do
    belongs_to :site, Repro.Site do
      allow_nil? false
      primary_key? true
    end
  end

  postgres do
    table "as"
    repo Repro.Repo
  end
end

defmodule Repro.Junction do
  use Ash.Resource, domain: nil, data_layer: AshPostgres.DataLayer

  actions do
    defaults([:create, :read, :update, :destroy])
  end

  attributes do
[...]

PR #848, opened 2026-09-10
fix: order no_phase operations after the objects they act on

MayCXC:
## Problem

An operation flagged `no_phase: true` carries no implicit ordering, so without a `requires/1` clause it floats ahead of the phased operations that build a table's structure. Two such operations render SQL against an object another operation creates in the *same batch*, so floating puts them before that object exists:

- **`AlterDeferrability`** (`ALTER CONSTRAINT ... DEFERRABLE`) runs against a foreign key an `AddAttribute` carrying `references:` creates. When a composite `with:` key pushes the foreign key into a later phase, the deferrability alter ran first and failed with `constraint ... does not exist`.
- **`AddPrimaryKey`** (`ADD PRIMARY KEY (keys)`) runs against key columns an `AddAttribute` can add in the same batch. Widening a primary key to include a newly added column (list-partitioning by that column, for instance) generated `ADD PRIMARY KEY (id, cell_id)` before `cell_id` existed.

## Fix

Give each a `requires/1` clause: `AlterDeferrability{direction: :up}` requires `table_columns_settled`; `AddPrimaryKey` requires `column_ready` for each key column. Both requirements are vacuous when the object predates the batch (nothing in the batch provides the fact), and neither adds a cycle: each provides only `table_structure_ready`, which `AddAttribute` never requires.

This builds on the dependency-graph model from #798, and is a distinct case from the composite-primary-key work in #805, which addressed the `streamline` merge and the `DropForeignKey{direction: :down}` phase split rather than `ADD PRIMARY KEY` over a newly added column.

## The class

The module doc now documents the whole class, so future `no_phase` operations are held to it. The rest are already safe: they act on objects that predate the batch (`RenameTable`, `MoveTableSchema`, `RemoveCustomIndex`, and the other renames and drops), render nothing in `up` (the `*Down` helpers), or are ordered by the consumers that require their facts.

## Tests

Reproducing tests for both fixes in `operation_deps_test.exs`, at both the `requires`/`provides` level and the `toposort_operations/1` level. Each fails without its `requires` clause (verified by neutralizing the clause).

- Full test suite passes.
- `mix ash_postgres.generate_migrations --check` is clean for both PG variants, so no checked-in migration reorders.

### DIFF
--- a/lib/migration_generator/operation_deps.ex
+++ b/lib/migration_generator/operation_deps.ex
@@ -19,20 +19,12 @@ defmodule AshPostgres.MigrationGenerator.OperationDeps do
   barrier would need it to run last — contradictory. Give each operation
   type the ordering it needs via `requires/1` instead.
 
-  Operations flagged `no_phase: true` are not grouped into the create/alter
-  phases that order the rest (see `group_into_phases/3`); they carry no
-  implicit ordering and float freely unless `requires/1` gives them an edge.
-  Any such operation whose `up` SQL names an object another operation can
-  create in the same batch (a column, a foreign key, a table, an index) has to
-  require that object's creation fact, or it can run before the object exists.
-  Most need no such requirement: they act only on objects that predate the
-  batch (a rename or drop of something an earlier migration made, e.g.
-  `RenameTable`, `RemoveCustomIndex`), render nothing in `up` (the `*Down`
-  helpers, which exist only to shape the reversed `down`), or are held in place
-  by the consumers that require their facts. The two that act on an
-  in-batch-created object are `AlterDeferrability` (the foreign key a new
-  `belongs_to` adds) and `AddPrimaryKey` (a column a new attribute adds); each
-  requires the relevant fact in `requires/1` below.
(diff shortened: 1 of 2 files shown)


lib/migration_generator/operation_deps.ex
```diff
@@ -19,6 +19,21 @@ defmodule AshPostgres.MigrationGenerator.OperationDeps do
   barrier would need it to run last — contradictory. Give each operation
   type the ordering it needs via `requires/1` instead.
 
+  Operations flagged `no_phase: true` are not grouped into the create/alter
```
zachdaniel: I think a lot of the comments/docs in this PR can be trimmed down *significantly*. Would you mind making that change? Otherwise LGTM

MayCXC: sure https://github.com/ash-project/ash_postgres/commit/092c37cb20138b7a727be811133dc781d5134802

zachdaniel (approved):

zachdaniel:

PR #858, opened 2026-09-19
test: cover two ash_sql sort bugs

jeroen11dijk:
These tests show two sort bugs in ash_sql. The fix is in https://github.com/ash-project/ash_sql/pull/263. Three of the five tests fail without the fix.

The other two tests pass with the fix and without it. This is correct. They make sure that the fix does not remove all sorts. One commit contains all the tests, because post.ex has changes for both tests.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #864, opened 2026-09-22
fix: convert Ecto.Query.CastError on every query path

### LINKED ISSUE #855: Ecto.Query.CastError is converted only where a rescue exists, and Ecto.SubQueryError is never unwrapped

## Request

With a UUID primary key and `query = Ash.Query.filter(Post, id == "not-a-uuid")`, `Ash.read(query)` returns `{:error, %Ash.Error.Invalid{errors: [%Ash.Error.Query.InvalidFilterValue{}]}}`. These do not:

| Case | Call | Result |
|---|---|---|
| a | `Ash.count(query)`, `Ash.exists(query)`, `Ash.aggregate(query, {:m, :max, field: :inserted_at})` | raises `Ecto.Query.CastError` |
| b | `query \|> Ash.Query.limit(1) \|> Ash.count()` | raises `Ecto.SubQueryError` |
| c | `Post \|> Ash.Query.load(comments: Comment \|> Ash.Query.filter(id == "not-a-uuid") \|> Ash.Query.limit(1)) \|> Ash.read()` with two or more posts | raises `Ash.Error.Unknown` wrapping `Ecto.SubQueryError` |
| d | `Post \|> Ash.Query.filter_input(%{comments: %{id: %{eq: "not-a-uuid"}}}) \|> Ash.read(page: [limit: 5])` | `{:error, %Ash.Error.Unknown{}}` wrapping `Ecto.SubQueryError` |
| e | AshJsonApi `PATCH /posts/not-a-uuid` | 500; `Ash.Error.Unknown` wrapping `Ecto.SubQueryError` from `update_query/4` |

With a single parent, case c converts cleanly, and `Ash.bulk_update/4` filtered by the bad id converts cleanly too (the error is a bare `CastError` there).

## Observed

```
** (Ecto.Query.CastError) deps/ash_sql/lib/expr.ex:1960: value `"not-a-uuid"` in `where` cannot be cast to type #Ash.Type.UUID.EctoType<[]> in query: ...
```
and, for the subquery cases:
```
** (Ecto.SubQueryError) the following exception happened when compiling a subquery.
    ** (Ecto.Query.CastError) ... value `"not-a-uuid"` in `where` cannot be cast ...
```
The `Ash.DataLayer` behaviour declares `{:ok, _} | {:error, _}` for `run_query`, `run_aggregate_query`, `run_aggregate_query_with_lateral_join` and `run_query_with_lateral_join`, so cases a to c also violate the callback contract. Downstream, ash_lua `operation = "count"` and `"exists"` return the raw cast error text and AshGraphql answers "Something went wrong" for the c and d shapes; those are effects of this bug, not separate ones.

## Expected

`{:error, %Ash.Error.Invalid{errors: [%Ash.Error.Query.InvalidFilterValue{}]}}` in every case, as on a plain read.

## Failure point

[lib/data_layer.ex](https://github.com/ash-project/ash_postgres/blob/v2.13.1/lib/data_layer.ex): `run_query/2` rescues at L925-L927; `run_aggregate_query/3` (L977), `run_query_with_lateral_join/4` (L1116, raise at L1148) and `run_aggregate_query_with_lateral_join/5` (L996, not exercised by a test) have no rescue; `handle_raised_error/4` converts a bare `Ecto.Query.CastError` at L3289 and sends `Ecto.SubQueryError` to the generic clause at L3324, from `run_query/2` (6d) and `update_query/4` (L1840, rescue at L1959, 6e).

## Notes

The GraphQL `{ count }` and JSON:API `page[count]=true` variants reach the count query only with `config :ash, :disable_async?, true`; the tests set it for those two cases. Related: an integer value outside the 64-bit range fails in the same function for a different exception, `DBConnection.EncodeError` (separate report). Also observed: the `InvalidFilterValue` produced by the working conversion carries the whole Ecto query as `context`, so `Exception.message/1` on it prints the query struct including `__ash_bindings__` with the actor and tenant.

## Reproduction

* Description, with the failure point at this release: https://github.com/grempe/ash-fuzz-repros#ash_postgresuncastable-filter-value-conversion
* Failing test: https://github.com/grempe/ash-fuzz-repros/blob/main/test/ash_postgres/uncastable_filter_value_conversion_test.exs
* Run: `mix setup && mix test test/ash_postgres/uncastable_filter_value_conversion_test.exs` in https://github.com/grempe/ash-fuzz-repros

grempe:
Closes #855.

Ash does not cast filter values against the attribute type (`Eq` declares `types: [:any, :same]`), so a value that cannot be cast reaches Ecto, which raises `Ecto.Query.CastError` while planning the query. `run_query/2` rescues that and converts it to `Ash.Error.Query.InvalidFilterValue`, but the conversion had two gaps, all with a UUID primary key and `id == "not-a-uuid"`:

- `run_aggregate_query/3`, `run_aggregate_query_with_lateral_join/5` and `run_query_with_lateral_join/4` have no rescue, so `Ash.count/2`, `Ash.exists/2`, `Ash.aggregate/3` and a limited relationship load across several parents raise the exception through the caller. The `Ash.DataLayer` behaviour declares `{:ok, _} | {:error, _}` for all four callbacks.
- `handle_raised_error/4` does not unwrap `Ecto.SubQueryError`, which Ecto raises around a cast error inside a subquery and which keeps the original in `exception`. So a count under a limit, a paginated read with a relationship filter, and `update_query/4` (AshJsonApi's `PATCH /posts/not-a-uuid`) return `Ash.Error.Unknown` even where a rescue exists.

Changes, all in `AshPostgres.DataLayer`:

- A `handle_raised_error/4` clause for `%Ecto.SubQueryError{exception: inner}` that re-dispatches the inner exception, so it converts the same way as at the top level.
- A `rescue` on the three callbacks, calling `handle_raised_error/4` with the query and resource the way `run_query/2` already does. `run_aggregate_query/3` delegates to `AshSql.AggregateQuery`; the rescue is placed here rather than in ash_sql because the other rescues and the conversion live here. `run_query_with_lateral_join/4` previously ignored its `destination_resource` argument; it is now used for the error.

Tests: `test/cast_error_test.exs` covers count, exists and aggregate (raised `CastError` on `main`), a count under a limit (raised `SubQueryError`), a limited relationship load across two parents and a relationship filter on a paginated read (both `Ash.Error.Unknown` on `main`), plus a control with a valid id on every path. Full suite 1006 passed on `main`, where the same three `JoinSubquerySortTest` / `UniqAggregateSortTest` cases from #858 fail with and without this change, and 983 passed with the patch on v2.13.1. `mix format --check-formatted`, `mix credo --strict`, `mix sobelow` and `mix dialyzer` clean.

**Overlap with #863.** That PR adds a clause to `handle_raised_error/4` at the same spot as the `SubQueryError` clause here, so whichever merges second will conflict on that one hunk. The resolution is to keep both clauses (order does not matter, they match different structs). I merged the two branches locally that way, and the full suite plus both new test files pass on the result. Happy to rebase this one once #863 lands, or the other way round.

Found by an AI agent working with a human fuzz-testing their own application; the reproduction outside this repo is https://github.com/grempe/ash-fuzz-repros/blob/main/test/ash_postgres/uncastable_filter_value_conversion_test.exs, which also covers the ash_lua, AshGraphql and AshJsonApi surfaces of the same paths.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

PR #865, opened 2026-09-23
test: range-valued temporal atomic writes

matt-beanland:
# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies

# Summary

This PR adds test coverage for a fix landing separately in `ash` core `temporal` branch, and will be red until that lands.

An atomic range update is resolving its target from the lower bound, not the record given.  A portion spanning two stored versions is refused instead of carving both; a portion entirely inside a version other than the one fetched targets that other version. A soft destroy has the same issue.

3 new tests added in one commit to `temporal_test`. Two added to existing `"atomics"` block, and one to a new `"soft destroy"` block, which required an action (`:cancel`, `soft?: true`) added to `Subscription`.

Full suite: 1046/1046 locally with the (as yet unraised) fix in place (1043 baseline + these 3), 2 failing without it.


matt-beanland:
the related ash fix PR is https://github.com/ash-project/ash/pull/2962

PR #866, opened 2026-09-23
fix: convert a character_not_in_repertoire error to an invalid-class error

### LINKED ISSUE #854: A NUL byte in text is an unconverted Postgrex.Error

## Request

A title containing a NUL byte (0x00): `Ash.create(Post, %{title: "a" <> <<0>> <> "b"})` and `Post |> Ash.Query.filter(title == "a" <> <<0>> <> "b") |> Ash.read()`. Through AshJsonApi: `POST /posts` with that title and `GET /posts?filter[title]=a%00b`. Through AshGraphql: `createPost` with the value as a variable or as a `\u0000` escape. Through ash_lua: `blog.post.create` with a Lua string literal containing the byte.

## Observed

The server rejects the value and the error is returned as unknown-class:
```
{:error, %Ash.Error.Unknown{errors: [%Ash.Error.Unknown.UnknownError{error:
  "** (Postgrex.Error) ERROR 22021 (character_not_in_repertoire) invalid byte sequence for encoding \"UTF8\": 0x00"}]}}
```
JSON:API answers 500; GraphQL answers "something went wrong"; Lua gets `unknown_error`. Invalid UTF-8 (`0xff`, reachable through a Lua literal) fails identically.

## Expected

An invalid-class error, as a constraint violation from the same server already gets (`Ash.Error.Changes.InvalidAttribute` for a create, `Ash.Error.Query.InvalidFilterValue` for a filter); a 400 on JSON:API. The Postgres error names no column, so attributing it to the attribute needs the changeset.

## Failure point

`handle_raised_error/4` in [lib/data_layer.ex](https://github.com/ash-project/ash_postgres/blob/v2.13.1/lib/data_layer.ex): the `Postgrex.Error` clauses at L3195, L3218, L3241, L3261, L3279 and L3298 all end in `to_constraints/2`, which yields nothing for `22021`; a create returns `to_ash_error/2` from L3269, a read from the catch-all clause.

## Notes

Postgrex maintainers consider input sanitising the application's job (elixir-ecto/postgrex#568, "Need way to sanitize string input", closed).

## Reproduction

* Description, with the failure point at this release: https://github.com/grempe/ash-fuzz-repros#ash_postgresnul-byte-in-text
* Failing test: https://github.com/grempe/ash-fuzz-repros/blob/main/test/ash_postgres/nul_byte_in_text_test.exs
* Run: `mix setup && mix test test/ash_postgres/nul_byte_in_text_test.exs` in https://github.com/grempe/ash-fuzz-repros

grempe:
Closes #854.

@g3nke11 claimed this on 2026-09-17. I linked this branch on the issue on 2026-09-22 and offered to open it if they had moved on; there has been no reply since, so I am opening it now. Happy to close it if they would rather finish their own.

PostgreSQL rejects text that is not valid UTF-8 or contains a NUL byte with `22021 character_not_in_repertoire`. Every `Postgrex.Error` clause in `handle_raised_error/4` ends in `Ecto.Adapters.Postgres.Connection.to_constraints/2`, which finds nothing for that code, so a create, update or filter with such a value returns `Ash.Error.Unknown`, and the API extensions answer 500. Only the data layer knows this storage has that limit; ETS and Mnesia store the value.

This adds a clause matching `%Postgrex.Error{postgres: %{code: :character_not_in_repertoire}}`. It matches on the error code, not on message text. It sits before the existing `Postgrex.Error` clauses for the `{:bulk_create, _}` and `{:ecto_changeset, _, _}` contexts, which would otherwise capture it first.

- In a query it becomes `Ash.Error.Query.InvalidFilterValue`.
- In an update it becomes `Ash.Error.Changes.InvalidAttribute` for each change whose value fails `String.valid?/1` or contains `<<0>>`. The error names no column, so the attribute comes from the changes.
- In a create it becomes `Ash.Error.Changes.InvalidChanges`. The rescue there only has the placeholder changeset that `bulk_create/3` builds for it, with no changes, so the attribute cannot be named.

The message is Postgres's own (`invalid byte sequence for encoding "UTF8": 0x00`). One thing I would change once the `no_value?` option you suggested on #863 exists: the query case has no value to put on `InvalidFilterValue`, so it currently renders as ``Invalid filter value `nil`: ...``, the same as #863. Happy to adjust it here along with that.

Tests: `test/repertoire_error_test.exs` covers a filter, an update with a NUL byte and with invalid UTF-8 (asserting the attribute), a create, and an ordinary-text control. On `main` the four error cases return `Ash.Error.Unknown`; with the change they pass. Full suite 1011 passed on `main`, where the same three `JoinSubquerySortTest` / `UniqAggregateSortTest` cases from #858 fail with and without this change, and 983 passed with the patch on v2.13.1. `mix format --check-formatted`, `mix credo --strict`, `mix sobelow` and `mix dialyzer` clean.

The branch is based on current `main`, and it also merges cleanly with #863 in either order; I checked `main` plus both together, and the full suite passes on that with only the three #858 failures.

Found by an AI agent working with a human fuzz-testing their own application; the reproduction outside this repo is https://github.com/grempe/ash-fuzz-repros/blob/main/test/ash_postgres/nul_byte_in_text_test.exs, where all eight cases pass against this branch, across Ash, AshJsonApi, AshGraphql and ash_lua.

# Contributor checklist

- [ ] Chores
- [ ] Documentation changes
- [ ] Refactoring
- [ ] Update dependencies


zachdaniel:

